PR #23, opened 2016-09-08
Add support for PROTO message value type.

marcushines:
This change allows the bytes to be a serialized proto message
If the sender uses this encoding it is up to the receiver to
Understand the schema and properly decode the message.

### DIFF
--- a/rpc/openconfig/openconfig.proto
+++ b/rpc/openconfig/openconfig.proto
@@ -230,6 +230,7 @@ message Value {
 enum Type {
   JSON = 0;
   BYTES = 1;
+  PROTO = 2;
 }
 
 // GetModelsRequest contains a list of models to return.  If no queries are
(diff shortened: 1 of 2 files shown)


aashaikh:
I think this is a good addition, but can we make this change along with the overall revisit of the message types?

marcushines:
Small incremental changes ...

On Sep 8, 2016 12:11 PM, "Anees Shaikh" <notifications@github.com> wrote:

I think this is a good addition, but can we make this change along with the
overall revisit of the message types?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/openconfig/reference/pull/23#issuecomment-245707336>,
or mute the thread

robshakir:
I don't think that this commit conflicts with other changes that we might want to make here. Suggest that we should merge this change and use something like release tags (or a version number within the proto) to indicate when changes are expected to be stable for implementors.

Merging the individual commits gives us the ability to trace changes back to why more easily IMHO.

aashaikh:
in that case, LGTM

tsuna:
When changing the `.proto` can you make sure to also submit the corresponding changes to the `.pb.go`?

Issue #36, opened 2016-11-22
gNMI Specification: using Unimplemented instead of InvalidArgument for unsupported encodings


tsuna:
This is a bit of a detail but since the spec specifies explicitly what to do in this case...  From Section 3.3.1 (The GetRequest Message):

> If the target does not support the specified encoding, the target MUST populate the error field of the `GetResponse` message, specifying an error of `InvalidArgument`.

`InvalidArgument` is for malformed input, but here if a client requests `JSON_IETF` encoding and the target doesn't support it, the request isn't malformed, it's just asking for something that is not supported by the target's implementation.

From https://godoc.org/google.golang.org/grpc/codes#Code

```go
    // InvalidArgument indicates client specified an invalid argument.
    // Note that this differs from FailedPrecondition. It indicates arguments
    // that are problematic regardless of the state of the system
    // (e.g., a malformed file name).
    InvalidArgument Code = 3
```
vs

```go
    // Unimplemented indicates operation is not implemented or not
    // supported/enabled in this service.
    Unimplemented Code = 12
```

There may be other instances of this issue in the spec.

tsuna:
In the same vein, I find this (in Section 3.4.7 [Error Handling]) a bit weird:

> When the operation is an update or replace operation that corresponds to a path that is not valid - `NotFound (5)`.

This seems inconsistent with Section 3.4.4 (Modes of Update: Replace versus Update):

> For both replace and update operations, if the path specified does not exist, the target MUST create the data tree element and populate it with the data in the `Update` message

If the path is not valid, I would imagine that `InvalidArgument` should be returned instead of `NotFound`.

tsuna:
Another instance of the problem mentioned in my first message is in Section 3.4.7 as well:

> _[...]_ `InvalidArgument (3)`. This error SHOULD be used _[...]_ in the case that multiple values are specified using a particular encoding (e.g., JSON) which cannot be decoded by the target.

Here I believe `Unimplemented` should be used instead.

robshakir:
I tend to agree here - in both cases. Particularly, the `Unimplemented` error is more explicit, I'll send a PR to adjust the specification to state this.

robshakir:
Closing this issue as it was resolved some time ago.

PR #43, opened 2017-03-08
gNMI 0.2.2

robshakir:
This PR requests to merge changes for gNMI 0.2.2, particularly, this:

- Clarifications to the `ON_CHANGE` behaviour of the `Subscribe` RPC, such that it is clearly stated that an `ON_CHANGE` path's current value should be sent when the RPC is initiated.
- Addresses inconsistencies in the recommended error code as per #36.

The following other changes are made:

- Regenerate Go stub with `protoc` v3.2.0.
- Generate gRPC stub for Python, and update the generated proto.
- Remove the PDF version of the specification.


rpc/gnmi/gnmi-specification.md
```diff
@@ -688,13 +680,12 @@ In the case that any operation within the `SetRequest` message fails, then (as p
...
 *   When the operation is an update or replace operation that corresponds to a path that is not valid - `NotFound (5)`. In this case the `message` field of the `Error` message specified SHOULD specify human-readable text indicating the path that was invalid.
-*   When the operation is an update or replace operation that includes an invalid value within the `Update` message specified - `InvalidArgument (3)`. This error SHOULD be used in cases where the payload specifies scalar values that do not correspond to the correct schema type, and in the case that multiple values are specified using a particular encoding (e.g., JSON) which cannot be decoded by the target.
+*   When the operation is an update or replace operation that includes an invalid value within the `Update` message specified - `InvalidArgument (3)`. This error SHOULD be used in cases where the payload specifies scalar values that do not correspond to the correct schema type,
+* When the client specifies a payload utilising an encoding that is not supported by the target (e.g., JSON) - `Unimplemented (12)` SHOULD be used to indicate that the encoding is not supported.
```
aashaikh: this could be a bit confusing -- wouldn't Unimplemented be more likely to be interpreted as the path not being supported?  InvalidArgument might be a better choice (i.e., the encoding is invalid, esp. if Capability has been used, but even if not)

rpc/gnmi/gnmi-specification.md
```diff
@@ -793,9 +776,9 @@ Stream subscriptions are long-lived subscriptions which continue to transmit upd
 
 A `STREAM` subscription is created by sending a `SubscribeRequest` message with the subscribe field containing a `SubscriptionList` message with the type specified as `STREAM`. Each entry within the `Subscription` message is specified with one of the following `modes`:
 
-
-
-*   **On Change (`ON_CHANGE`)** - when a subscription is defined to be "on change", data updates are only sent when the value of the data item changes. A heartbeat interval MAY be specified along with an "on change" subscription - in this case, the value of the data item(s) MUST be re-sent once per heartbeat interval regardless of whether the value has changed or not.
+*   **On Change (`ON_CHANGE`)** - when a subscription is defined to be "on change", data updates are only sent when the value of the data item changes.
+    * For all `ON_CHANGE` subscriptions, the target MUST generate updates for all path that match the subscription path(s), and transmit them. Following this initial set of updates, updated values SHOULD only be transmitted when their value changes.
```
aashaikh: ... for all *paths* that match ...

maybe clarify that the target must *first* generate updates for all paths as with any other subscription

rpc/gnmi/gnmi-specification.md
robshakir: I don't think that this is confusing - it's a valid argument (`InvalidArgument` would seem to say that the path is not valid - per the above). `Unimplemented` seems to me to be clearer for the encoding - we don't use `Unimplemented` elsewhere either.

rpc/gnmi/gnmi-specification.md
robshakir: Done.

robshakir:
@aashaikh PTAL, would like to merge this. Apologies for the delay in getting back to you.

rpc/gnmi/gnmi-specification.md
aashaikh: ok,  I still think I would have a hard time interpreting this to mean that the encoding is not supported but I don't think there is a strictly better option.

PR #45, opened 2017-04-05
gNMI 0.3.0: Add native protobuf encoding support

robshakir:
This PR requests to merge a number of changes that make up v0.3.0 of gNMI. The focus of these changes is to support native protobuf types as the encoding for values within gNMI `Update` messages. The intention is to allow implementations to serialise directly to protobuf (rather than marshalling to JSON), such that encoding ambiguity is reduced. Explicit support is also included for marshalling an update to `protobuf.Any` for implementations which natively support protobuf encoding for elements in their schema.

The particular changes are:
- **Added a `oneof val` to the `Update` message**: This changes the method of determining the type of data that is encoded within an `Update` from using the `Encoding` enumeration, to determining which of the fields within the `val` `oneof` are populated. The `oneof` includes fields corresponding to native protobuf types, along with fields that contain data of one of these types which is encoded according to a specific set of rules (e.g., `json`, `json_ietf`, etc.).
- **Deprecated the `Value` message, and `Update` `value` field**: In order to remove the duplication that otherwise the `val` `oneof` would introduce, deprecate the former method of encoding a value. The motivation for this change in encoding is improved efficiency of the resulting encoding.
- **Add new `Decimal64`, `ScalarArray` and `ScalarArrayElement` messages**: These messages allow for specific YANG-modeled data types to be encoded within the `Update` message's `val` field. `Decimal64` specifically handles YANG's `decimal64` value, `ScalarArray` and `ScalarArrayValue` allow encoding of `leaf-list` schema elements that may be mixed type.
- **Add a new `encoding` field to `SubscriptionList`**: This allows a client to specify to a target how the data that is returned within `Notification` messages for a particular subscription should be encoded.

In addition to the proto change, the generated code for both Go and Python has been updated.


rpc/gnmi/gnmi.proto
```diff
@@ -129,6 +149,36 @@ message Error {
...
+  // One of the fields within the value oneof may be specified
+  // indicating an element included within the ScalarArray.
+  oneof value {
+    string string_val = 1;      // String value.
+    int64 int_val = 2;          // Integer value.
+    uint64 uint_val = 3;        // Unsigned integer value.
+    bool bool_val = 4;          // Boolean value.
+    bytes bytes_val = 5;        // Arbitrary byte sequence value.
+    float float_val = 6;        // Floating-point value.
+    Decimal64 decimal_val = 7;  // Decimal value.
```
pborman: Should there be a ScalarArray field?

robshakir: In YANG we can't have a leaf-list that contains a leaf-list so I omitted this case.

rpc/gnmi/gnmi.proto
```diff
@@ -101,9 +120,10 @@ message Path {
 }
 
 // Value encodes a data tree node's value - along with the way in which
-// the value is encoded.
+// the value is encoded. This field is deprecated by gNMI 0.3.0.
```
robshakir: Done, thanks.

rpc/gnmi/gnmi.proto
```diff
@@ -87,8 +87,27 @@ message Notification {
...
+    bool bool_val = 13;               // Bool value.
+    bytes bytes_val = 14;             // Arbitrary byte sequence value.
+    float float_val = 15;             // Floating point value.
+    Decimal64 decimal_val = 16;	      // Decimal64 encoded value.
+    ScalarArray leaflist_val = 17;    // Mixed type scalar array value.
+    google.protobuf.Any any_val = 18; // protobuf.Any encoded bytes.
+    string json_val = 19;             // JSON-encoded text.
+    string json_ietf_val = 20;        // JSON-encoded text per RFC7951.
+    string ascii_val = 21;            // Arbitrary ASCII text.
+  }
```
robshakir: Thanks, done.

PR #47, opened 2017-04-05
gNMI 0.3.0.

robshakir:
(NB: this PR replaces #45 for further review)

This PR requests to merge a number of changes that make up v0.3.0 of gNMI. The focus of these changes is to support native protobuf types as the encoding for values within gNMI `Update` messages. The intention is to allow implementations to serialise directly to protobuf (rather than marshalling to JSON), such that encodiang ambiguity is reduced. Explicit support is also included for marshalling an update to `protobuf.Any` for implementations which natively support protobuf encoding for elements in their schema.

The particular changes are:
- **Added a `oneof val` to the `Update` message**: This changes the method of determining the type of data that is encoded within an `Update` from using the `Encoding` enumeration, to determining which of the fields within the `val` `oneof` are populated. The `oneof` includes fields corresponding to native protobuf types, along with fields that contain data of one of these types which is encoded according to a specific set of rules (e.g., `json`, `json_ietf`, etc.).
- **Deprecated the `Value` message, and `Update` `value` field**: In order to remove the duplication that otherwise the `val` `oneof` would introduce, deprecate the former method of encoding a value. The motivation for this change in encoding is improved efficiency of the resulting encoding.
- **Add new `Decimal64`, `ScalarArray` and `ScalarArrayElement` messages**: These messages allow for specific YANG-modeled data types to be encoded within the `Update` message's `val` field. `Decimal64` specifically handles YANG's `decimal64` value, `ScalarArray` and `ScalarArrayValue` allow encoding of `leaf-list` schema elements that may be mixed type.
- **Add a new `encoding` field to `SubscriptionList`**: This allows a client to specify to a target how the data that is returned within `Notification` messages for a particular subscription should be encoded.

In addition to the proto change, the generated code for both Go and Python has been updated.


rpc/gnmi/gnmi.proto
```diff
@@ -87,8 +87,27 @@ message Notification {
...
-  Path path = 1;    // The path (key) for the update.
-  Value value = 2;  // The value (value) for the update.
+  Path path = 1;                      // The path (key) for the update.
+  Value value = 2 [deprecated=true];  // The value (value) for the update.
+  // One of the fields within the val oneof is populated with the value
+  // of the update. The type of the value being included in the Update
+  // determines which field should be populated. In the case that the
+  // encoding is a particular form of the base protobuf type, a specific
+  // field is used to store the value (e.g., json_val).
+  oneof val {
```
gcsl: One advantage of having an explicit message type for the value containing the oneof is that it can be referenced in other places. I am leaning in favor of having a single message type with the oneof that can be used both for the value and the leaflist with comments clearly identifying which elements are not eligible to be leaflist members.

robshakir: I was a bit torn between the two approaches, as we discussed. PTAL at the refactored version that I just pushed.

robshakir:
@gcsl I didn't want to list YANG-specific information in the proto. There is no current link between YANG and gNMI, and I don't want to create one. I can simply remove this comment if we think it adds no value.

rpc/gnmi/gnmi.proto
```diff
@@ -129,6 +155,23 @@ message Error {
   google.protobuf.Any data = 3;   // Optional additional information.
 }
 
+
+// Decimal64 is used to encode floating-point decimal number. The value
+// expressed as a set of digits with the precision specifying the number
```
aashaikh: ... value is expressed ...

rpc/gnmi/gnmi.proto
```diff
@@ -129,6 +155,23 @@ message Error {
...
+  uint64 digits = 1;      // Set of digits.
+  uint32 precision = 2;    // Number of digits following the decimal point.
+}
+
+// ScalarArray is used to encode a mixed-type array of values.
+message ScalarArray {
+  // The set of elements within the array. Each TypedValue message should
+  // specify only elements that have a field identifier of 1-7 (i.e., the
+  // values are scalar values).
+  repeated TypedValue element = 1;
```
aashaikh: given the restriction on the types in the scalar array, does it make sense to list out the types here, or maybe split TypedValue message into 'simple' and 'complex' type messages ?

robshakir: Splitting into different messages has some usability issues of causing another level of messages in the generated code (there is already one more due to the `oneof`). I think we should leave this as is. I think just adding field IDs here feels sufficient.

rpc/gnmi/gnmi.proto
robshakir: Done, thanks.

rpc/gnmi/gnmi.proto
aashaikh: yes, I didn't think enough about the pain to populate this -- SGTM.

PR #48, opened 2017-04-20
gNMI 0.3.1

robshakir:
```
    Update SetResponse error and timestamp.
    
      * (M) rpc/gnmi/gnmi.proto
        - SetResponse utilised an Error message, which duplicates the
          contents of google.rpc.Status when returned. This change
          deprecates the top-level Error, preferring to use rpc.Status.
          If additional details are to be included by the target when
          returning details to the server, then the google.protobuf.Any
          field of rpc.Status can be used to store a gnmi.Error message.
          This change reduces complexity at the client which otherwise
          must handle the case that the error field indicates an
          incomplete Set RPC, but the grpc.Code in the return indicates
          success.
        - Deprecate the timestamp field of the UpdateResult message.
          gNMI defines all operations within a Set RPC to be a single
          transaction, therefore it is ambiguous what this timestamp
          implies - since validation can only be performed at the end
          of the transaction.
        - Add a new timestamp field to the SetResponse message to
          timestamp the overall transaction.
```


rpc/gnmi/gnmi.proto
```diff
@@ -304,14 +304,18 @@ message SetRequest {
 
 // SetResponse is the response to a SetRequest, sent from the target to the
 // client. It reports the result of the modifications to the data tree that were
-// specified by the client.
+// specified by the client. Errors for this RPC should be reported using the
+// https://github.com/googleapis/googleapis/blob/master/google/rpc/status.proto
+// message in the RPC return. The gnmi.Error message can be used to add additional
```
aashaikh: maybe just clarify that `gnmi.Error` can be included via the protobuf.Any `details`  field

rpc/gnmi/gnmi.proto
```diff
@@ -325,7 +329,11 @@ message UpdateResult {
     REPLACE = 2;          // The result relates to a replace of Path.
     UPDATE = 3;           // The result relates to an update of Path.
   }
-  int64 timestamp = 1;    // Timestamp of generation (ns since epoch).
+  // Deprecated timestamp for the UpdateResult, this field has been
+  // replaced by the timestamp within the SetResponse message, since
+  // all mutations effected by a set should be applied as a single
+  // transaction.
+  int64 timestamp = 1 [deprecated=true];
```
aashaikh: Agree with adding a timestamp for the overall Set transaction, but is it problematic to leave this field as-is instead of deprecating?  It could provide a way to return timestamps to indicate completion of individual operations within the transaction (optionally, for targets that can support it).

brbuffam:
When can we expect the corresponding update of the specification?

aashaikh:
we'll plan to get the spec doc updated in the next week or so.

PR #49, opened 2017-04-25
Add certificate installation and removal from the target

marcushines:


rpc/gnoi/gnoi_cert.proto
```diff
@@ -0,0 +1,378 @@
...
+  // in the Rotate() RPC.
+  //
+  // Sequence of expected messages:
+  // Case 1: When Target generates the CSR.
+  //
+  //   Step 1: Start the stream
+  //     Client <---- Rotate() RPC stream begin ------> Target
+  //
+  //   Step 2: CSR
+  //     Client -----> GenerateCSRRequest() ----> Target
```
robshakir: Should these have `()` after them, since they seem to be the submessages of the `RotateCertificateRequest` rahter than RPC which they seem to read like right now?

rpc/gnoi/gnoi_cert.proto
```diff
@@ -0,0 +1,378 @@
...
+  //     Client <----- GenerateCSRResponse() <--- Target
+  //
+  //   Step 3: Certificate Signing
+  //     Client gets the certificate signed by the CA.
+  //
+  //   Step 4: Send Certificate to Target.
+  //     Client --> LoadCertificateRequest() ----> Target
+  //     Client <-- LoadCertificateResponse() <--- Target
+  //
+  //   Step 5: Test/Validation by the client.
```
robshakir: What test/validation do we expect here? Does the documentation need to reflect this?

rpc/gnoi/gnoi_cert.proto
```diff
@@ -0,0 +1,378 @@
...
+  //
+  //   Step 6: Final commit.
+  //     Client ---> FinalizeRequest() ----> Target
+  rpc Rotate(stream RotateCertificateRequest)
+      returns (stream RotateCertificateResponse);
+
+  // Install will put a new Certificate on the target by creating a new CSR
+  // request and placing the new Certificate based on the CSR on the target.The
```
robshakir: Maybe reword "new Certificate based on the CSR" to "corresponding certificate"?

rpc/gnoi/gnoi_cert.proto
```diff
@@ -0,0 +1,378 @@
...
+//
+// When this request is made for installing a completely new certificate as part
+// of the Install() RPC , then the target must ensure that the "certificate_id"
+// is completely new and no entities on the target are should be bound to this
+// certificate_id. If any existing certificate matches the certificate_id, then
+// this request should fail.
+//
+// If there is another ongoing Rotate/Install RPC with the same certificate_id,
+// the GenerateCSRRequest should fail.
+// should fail.
```
robshakir: Erroneous "should fail".

rpc/gnoi/gnoi_cert.proto
```diff
@@ -0,0 +1,378 @@
...
+}
+
+// Parameters to be used when generating a Certificate Signing Request.
+message CSRParams {
+  // The type of certificate which will be will be associated for this CSR.
+  CertificateType type = 1;
+
+  // Minimum size of the key to be used by the target when generating a
+  // public/private key pair.
+  uint32 min_key_size = 2;
```
robshakir: Should this rather be the actual key size that one wants to request? It seems strange to just give a minimum - for example, if I specify 1024 here, should the target generate a 2048 or 16384 sized key?

rpc/gnoi/gnoi_cert.proto
```diff
@@ -0,0 +1,378 @@
...
+message RevokeCertificatesRequest {
+  // Certificates to revoke.
+  repeated string certificate_id = 1;
+}
+
+message RevokeCertificatesResponse {
+  // List of certificates successfully revoked.
+  repeated string revoked_certificate_id = 1;
+
+  // List of errors why certain certificates could not be revoked.
```
robshakir: Should this just be included as a `proto.Any` in `status.Status` [https://github.com/googleapis/googleapis/blob/master/google/rpc/status.proto]?

marcushines: no as this is a "successful" response not an error status should only contain metadata in a result of error

rpc/gnoi/gnoi_cert.proto
marcushines: it can

rpc/gnoi/gnoi_cert.proto
marcushines: meh

rpc/gnoi/gnoi_cert.proto
marcushines: done

rpc/gnoi/gnoi_cert.proto
marcushines: done

rpc/gnoi/gnoi_cert.proto
marcushines: done

PR #80, opened 2017-11-21
Update gnmi specification to version 0.5.0, added target, duplicates and updates_only fields.

gcsl:


rpc/gnmi/gnmi-specification.md
```diff
@@ -182,6 +184,10 @@ path: <
 
 Paths (defined to be the concatenation of the `prefix` and `path` within the message) specified within a message MUST be absolute - no messages with relative paths should be generated.
 
+#### 2.2.2.1 Path Target
+
+The `target` field within a path is name for the target. This field MUST only ever be present on `prefix` paths in the corresponding request and response messages.  This field is optional for clients that when set in the `prefix` in a request, `GetRequest`, `SetRequest` or `SubscribeRequest`, MUST be reflected in the `prefix` field of the corresponding `GetResponse`, `SetResponse` or `SubscribeResponse` by a server. The intention of this field is to allow a name to be associated with all the data for a given stream if requested by a client. If a client does not set this field in the `prefix` of a request, it MUST NOT be set in the `prefix` of the corresponding response messages. Simlultaneous connections by multiple clients could be comprised of a mix of set and unset `target` fields, and the value could differ among simultaneous clients.
```
robshakir: This restriction does close off one thing - which is that a proxy cannot respond with a single response for >1 device. Is this something that we're happy with?

robshakir: "client that when set" -> "clients. When set"

robshakir: "The intention of this field is to" -> "This field is used to..."

robshakir: s/Simlultaneous/Simultaneous/ - but not sure whether we should have this last sentence. I'm not sure I quite understand what it is saying.

gcsl: With timestamps in nanoseconds, I think it unlikely that a proxy would be able to aggregate across devices anyway.

gcsl: Done

gcsl: Done

gcsl: The meaning is that the value is tied to the context of the client RPC and not shared among multiple clients.  Maybe we should just say that?

gcsl: New wording.

robshakir:
LGTM :ship:

PR #100, opened 2018-06-12
Add a document explaining mixed-schema in gNMI.

robshakir:
```
 * (A) reference/gnmi/mixed-schema.md
  - Add a reference document covering the use of mixed schema in
    gNMI. This covers use of the origin field for multiple schemas
    being used within gNMI RPCs.
```


robshakir:
Thanks!

PR #107, opened 2018-11-20
Add document covering how to represent gNMI paths as strings.

robshakir:
This PR adds a document which provides a specification for going from a gNMI `Path` message to a string, to ensure that there is a stable reference for implementations.

### DIFF
--- a/rpc/gnmi/gnmi-path-strings.md
+++ b/rpc/gnmi/gnmi-path-strings.md
@@ -1,0 +1,213 @@
+## Representing gNMI Paths as Strings
+
+**Authors**: 
+robjs<sup>⸸</sup>, aashaikh<sup>⸸</sup>, csl<sup>⸸</sup>, hines<sup>⸸</sup>.  
+<sup>⸸</sup> @google.com
+
+October 2018
+
+
+## Overview
+
+This document provides a specification for the representation of a gNMI
+[`Path`](https://github.com/openconfig/gnmi/blob/master/proto/gnmi/gnmi.proto#L133)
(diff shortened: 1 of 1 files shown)


robshakir:
@gcsl @aashaikh -- wondering whether one of you can LGTM this, based on previous discussions? Any other review comments are welcome. Thanks!

robshakir:
Thanks! :-)

rpc/gnmi/gnmi-path-strings.md
```diff
@@ -0,0 +1,213 @@
...
+map. Where multiple keys are present (as shown with `key1` and `key2`) the
+values are encoded in the form:
+
+```
+[key=value]
+```
+
+and appended to the string. If the key value contains a `]` character, it is
+escaped by preceding it with the backslash (`\`) character. Only `]` and `\`
+characters must be escaped. Newlines and carriage returns are encoded as `\n`
```
mkh-arista, line 40: Is `\` really escaped and unescaped in the reference implementation? It seems like only `\` followed by `]` is handled, and all other occurrences of `\` are left untouched. Which one is right?

rpc/gnmi/gnmi-path-strings.md
```diff
@@ -0,0 +1,213 @@
...
+and `\r` respectively.  Unicode characters are encoded as `\u1234` or `\U1234`.
```
mkh-arista, line 41: Are `\u1234` and `\U1234` equivalent, or should it be `\U12345678`?

mkh-arista (commented):
Hello, thank you very much for this document. Couple of clarifying questions below.

PR #118, opened 2020-03-11
Add example of set-replace

mike-albano:
Add example of set-replace semantics in the presence of defaults vs. non-defaults.

### DIFF
--- a/rpc/gnmi/gnmi-specification.md
+++ b/rpc/gnmi/gnmi-specification.md
@@ -1119,6 +1119,29 @@ omits values that have been previously set, they MUST be treated as deleted from
 the data tree.  Otherwise, omitted data elements MUST be created with their
 default values on the target.
 
+For example, consider the following tree, where boolean `b` has a default value
+of True and boolean `c` has no default value. Both `b` and `c` have been
+previously set as False.
+```
+root +
+     |
+     + a --+
+     |     |
+     |     +-- b
+     |     |
+     |     +-- c
+     |
+     |
+     + d --+
+           +-- e
+           |
+           +-- f
+```
+A `replace` operation is issued where `e` and `f` are set, and all other
+elements are omitted. In this scenario, `b` MUST be reverted to it's default
+setting of True and the configuration of `c` MUST be deleted from the tree,
+and returned to it's original un-configured setting.
+
 For `update` operations, only the value of those data elements that are
 specified explicitly should be treated as changed.
 
--- a/rpc/gnmi/gnmi-specification.md
+++ b/rpc/gnmi/gnmi-specification.md
@@ -1112,12 +1112,15 @@ updates within the `SetRequest` method, return the data tree to the state prior
 to any changes, and return a `SetResponse` status indicating the error
 encountered.
 
-For `replace` operations, the behavior regarding omitted data elements in the
-`Update` depends on whether they refer to non-default values (i.e., set by a
-previous `SetRequest`), or unmodified defaults.  When the `replace` operation
-omits values that have been previously set, they MUST be treated as deleted from
-the data tree.  Otherwise, omitted data elements MUST be created with their
-default values on the target.
+For replace operations:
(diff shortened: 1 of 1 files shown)


rpc/gnmi/gnmi-specification.md
```diff
@@ -1112,12 +1112,38 @@ updates within the `SetRequest` method, return the data tree to the state prior
...
-the data tree.  Otherwise, omitted data elements MUST be created with their
-default values on the target.
+For replace operations:
+
+ * If a particular path-value is specified in the client-supplied data, it is
+ replaced with the client-specified value.
+ * If a particular path-value is NOT specified in the client-supplied data and
+ the path does not have a specified default value in the corresponding schema,
+ it is treated as deleted.
+ * If a path-val is NOT specified in the client-supplied data and the path does
```
robshakir: s/path-val/path-value/

rpc/gnmi/gnmi-specification.md
```diff
@@ -1112,12 +1112,38 @@ updates within the `SetRequest` method, return the data tree to the state prior
...
+ have a default value, it takes on this value.
```
robshakir: s/this value/the schema-specified default value/

rpc/gnmi/gnmi-specification.md
```diff
@@ -1112,12 +1112,38 @@ updates within the `SetRequest` method, return the data tree to the state prior
...
+
+For example, consider the following tree, where boolean `b` has a default value
```
robshakir: where a schema defines the boolean value `b` as having a default value of true?

rpc/gnmi/gnmi-specification.md
```diff
@@ -1112,12 +1112,38 @@ updates within the `SetRequest` method, return the data tree to the state prior
...
+     |     +-- c
+     |
+     |
+     + d --+
+           +-- e
+           |
+           +-- f
+```
+A `replace` operation is issued where `e` and `f` are set, and all other
+elements are omitted. In this scenario, `b` MUST be reverted to it's default
```
robshakir: s/it's/its/

rpc/gnmi/gnmi-specification.md
mike-albano: Done

rpc/gnmi/gnmi-specification.md
mike-albano: Done

rpc/gnmi/gnmi-specification.md
mike-albano: Took a shot at slightly modified version of this suggestion.

rpc/gnmi/gnmi-specification.md
mike-albano: Done.

rpc/gnmi/gnmi-specification.md
```diff
@@ -1112,12 +1112,38 @@ updates within the `SetRequest` method, return the data tree to the state prior
...
-omits values that have been previously set, they MUST be treated as deleted from
-the data tree.  Otherwise, omitted data elements MUST be created with their
-default values on the target.
+For replace operations:
+
+ * If a particular path-value is specified in the client-supplied data, it is
+ replaced with the client-specified value.
+ * If a particular path-value is NOT specified in the client-supplied data and
+ the path does not have a specified default value in the corresponding schema,
+ it is treated as deleted.
```
aashaikh: should we just say "it should be deleted"

mike-albano: agreed, changed.

mike-albano: "should" or "MUST"?

aashaikh: I think you're right -- it should say MUST; otherwise vendors might do odd things like set it to some random value.

PR #127, opened 2020-06-12
gNMI specification improvements

hellt:
Hi all,
in this PR:

1. Table of contents format is aligned to left
2. Dead links in TOC are removed with the working ones
3. Several internal cross-section links fixed
4. a few html links `<a href>` changed to markdown style `[link](url)`


hellt:
Hi @robshakir 
is it the right approach to suggest minor improvements to the spec? Or is supposed to be driven from inside of OC group?

PR #128, opened 2020-08-04
Add gNMI History extension documentation

jcostaroberts:
This change adds the specification for the gNMI History extension. This extension lets clients request historical data.

### DIFF
--- a/rpc/gnmi/gnmi-history.md
+++ b/rpc/gnmi/gnmi-history.md
@@ -1,0 +1,87 @@
+# gNMI History Extension
+
+**Contributors:** Justin Costa-Roberts, Aaron Beitch, Mike Fink
+
+**Date:** August 4, 2020
+
+**Version:** 0.1.0
+
+# 1. Purpose
+
+In certain applications it is necessary not only to retrieve current data from a
+data tree, but also to allow querying historical data. The gNMI History
+extension enables clients both to retrieve data at a specific time in the past,
+and to request all updates applied to a data tree between two specified times.
+
+# 2. Definition
+
+A `History` message is embedded in an Extension message in the
+`SubscribeRequest` proto.
+
+## 2.1 Proto
+
+```
+message TimeRange {
+  int64 start = 1;
+  int64 end = 2;
+}
+
+message History {
+  oneof request {
+    int64 snapshot_time = 1;
+    TimeRange range = 2;
+  }
+}
+```
+
+## 2.2 SubscribeRequest handling
+
+The History extension may be applied only to `ONCE` and `STREAM` subscriptions.
+Requests for historical data can be either of two types: a snapshot at a
+specified time, or a set of updates between two times. The `History` message's
+request field is a `oneof` field whose type indicates which of these request
+types applies.
+
+In all cases, times are represented as int64s interpreted as nanoseconds since
+the Unix Epoch.
+
+### 2.2.1 Snapshot requests
+
+If the `snapshot_time` field is set, the request is for a snapshot. The
+subscription mode for a snapshot request must be `ONCE`; otherwise the server
+must return an `INVALID_ARGUMENT` error. The server must transmit the data
+indicated in the `SubscribeRequest` as it existed on the server at the specified
+time. If the specified time indicates a future time at the point the server
+receives the request, the server may either return an `UNIMPLEMENTED` error or
+choose to wait until the specified time has been reached.
+
+### 2.2.2 Range requests
+
+If the `range` field is set, the request is for the set of updates that occurred
+during the half-open interval between the specified start and end times. Such a
+request may return multiple updates per leaf, so the subscription mode for a
+range request must be `STREAM`. If the mode of a range request is not `STREAM`,
+the server should return an `INVALID_ARGUMENT` error. For the paths specified in
+the `SubscriptionList`, the server must transmit all updates it's aware of that
+occurred at the start time or later and before the end time specified in the
+`TimeRange`, and then close the RPC. As usual, the server should also transmit
+the specified data tree as it existed at the subscription start time unless the
+`updates_only` field on the `SubscriptionList` message is set.
+
+If `TimeRange`'s start time is later than its end time, the server must return
+an `INVALID_ARGUMENT` error. If a start or end time has not yet elapsed when the
+server receives the request, the server may either return an `UNIMPLEMENTED`
+error or honor the request by waiting until the specified time. Clients may
+therefore request open-ended `STREAM` subscriptions beginning in the past by
+setting the end time to, for example, the largest int64 value.
+
+### 2.2.3 Considerations for requests not immediately serviceable
+
+Clients should not expect servers to service requests with start or end times in
+the future. Servers may honor such requests in order to account for the
+possibility of moderate differences in local time between the client and server,
+but clients should expect that times far into the future will not be serviced,
+and that significant clock desynchronization may cause the server to return an
+error for requests with specified times near the present. Finally, clients
+should expect that in order not to exhaust resources, servers may limit the
+number of open requests that cannot immediately be serviced.


jcostaroberts:
@googlebot I signed it!

jcostaroberts:
@googlebot I signed it!

rpc/gnmi/gnmi-history.md
```diff
@@ -0,0 +1,47 @@
+# gNMI History Extension
```
robshakir, line 1: (nit): Can you 80ch wrap the doc by any chance? It just makes things easier to review in the future.

rpc/gnmi/gnmi-history.md
```diff
@@ -0,0 +1,47 @@
...
+
+## 2.2 SubscribeRequest handling
+
+The History extension may be applied only to `ONCE` and `STREAM` subscriptions. Requests for historical data can be either of two types: a snapshot at a specified time, or a set of updates between two times. The `History` message's request field is a `oneof` field whose type indicates which of these request types applies.
+
+In all cases, times are represented as int64s interpreted as nanoseconds since the Unix Epoch.
+
+### 2.2.1 Snapshot requests
+
+If the `time` field is set, the request is for a snapshot. The subscription mode for a time request must be `ONCE`; otherwise the server must return an `INVALID_ARGUMENT` error. The server must transmit the data indicated in the `SubscribeRequest` as it existed on the server at the specified time. If the specified time indicates a future time at the point the server receives the request, the server may either return an `UNIMPLEMENTED` error or choose to wait until the specified time has been reached.
```
robshakir, line 41: I wonder whether it's a good idea to wait until the specified time has been reached -- it seems like a potential DoS vector if a client can send a lot of requests that potentially are held open for a long time. Is there a specific use case that motivates this?

rpc/gnmi/gnmi-history.md
```diff
@@ -0,0 +1,47 @@
...
+
+If the `range` field is set, the request is for the set of updates that occurred between the specified start and end times. Such a request may return multiple updates per leaf, so the subscription mode for a range request must be `STREAM`. If the mode of a range request is not `STREAM`, the server should return an `INVALID_ARGUMENT` error. For the paths specified in the `SubscriptionList`, the server must transmit all updates it's aware of that occurred at the start time or later and before the end time specified in the `TimeRange`, and then return. Whether the server should transmit the specified data tree as it existed at the start time is determined as usual by the `updates_only` field on the `SubscriptionList` message.
+
+If `TimeRange`'s start time is later than its end time, the server must return an `INVALID_ARGUMENT` error. If a start or end time has not yet elapsed when the server receives the request, the server may either return an `UNIMPLEMENTED` error or honor the request by waiting until the specified time. Clients may therefore request open-ended `STREAM` subscriptions beginning in the past by setting the end time to, for example, the largest int64 value.
```
robshakir, line 47: Similar question here as to whether a start time in the future should be supported.

robshakir (commented):
I have one clarification question here -- but this LGTM.

@mkhsueh @gscl -- PTAL and see whether you're good with this.

rpc/gnmi/gnmi-history.md
```diff
@@ -0,0 +1,47 @@
...
+
+# 2.2.2 Stream requests
+
+If the `range` field is set, the request is for the set of updates that occurred between the specified start and end times. Such a request may return multiple updates per leaf, so the subscription mode for a range request must be `STREAM`. If the mode of a range request is not `STREAM`, the server should return an `INVALID_ARGUMENT` error. For the paths specified in the `SubscriptionList`, the server must transmit all updates it's aware of that occurred at the start time or later and before the end time specified in the `TimeRange`, and then return. Whether the server should transmit the specified data tree as it existed at the start time is determined as usual by the `updates_only` field on the `SubscriptionList` message.
```
mkhsueh, line 45: (nit): Consider calling out the "between" behavior as the half-open interval, perhaps at the start of this paragraph; the spec mentions it thrice before defining semantics.

rpc/gnmi/gnmi-history.md
jcostaroberts: We went back and forth on this. The reason to wait is that it seems undesirable for a client request ending at t=time.Now() to result in an error just because the server's time is slightly behind the client's. Another approach we considered is to have the server immediately return all updates between the specified start time and its own current time u (u < t). This also seems undesirable because the client might well use time t as the start time for a later request (if, say, it does a get every hour), which would create a window between u and t for which it would not receive updates, and wouldn't know it hadn't.

The approach we landed on was either to give the client exactly what it asked for or to return an error, and to leave it up to the server whether it wants to wait. In practice I imagine a server would have some short period it would be willing to wait, and otherwise would return an error.

What do you think?

rpc/gnmi/gnmi-history.md
jcostaroberts: Done.

rpc/gnmi/gnmi-history.md
jcostaroberts: Done.

rpc/gnmi/gnmi-history.md
robshakir: That sounds reasonable to me -- why don't we add a small note here that just explains that a client should expect that there are conditions in which:

* Connections with start times that that are far into the future may be rejected.
* Servers may limit the number of connections of this nature.

And notes that server implementors should be careful not to cause themselves a resource exhaustion problem here. It's stepping beyond a spec into implementation advice, but I think we did something similar in the 'Considerations for using Get' section of the main spec.

jcostaroberts: Yeah that seems good. I've added a section at the end of the document including these points.

rpc/gnmi/gnmi-history.md
```diff
@@ -0,0 +1,87 @@
...
+
+If the `time` field is set, the request is for a snapshot. The subscription mode
+for a time request must be `ONCE`; otherwise the server must return an
+`INVALID_ARGUMENT` error. The server must transmit the data indicated in the
+`SubscribeRequest` as it existed on the server at the specified time. If the
+specified time indicates a future time at the point the server receives the
+request, the server may either return an `UNIMPLEMENTED` error or choose to wait
+until the specified time has been reached.
+
+### 2.2.2 Stream requests
```
mhines01, line 58: Does the stream include an initial snapshot? or only updates received between the date ranges?

rpc/gnmi/gnmi-history.md
```diff
@@ -0,0 +1,87 @@
...
+
+```
+message TimeRange {
+  int64 start = 1;
+  int64 end = 2;
+}
+
+message History {
+  oneof request {
+    int64 time = 1;
```
mhines01, line 31:
```proto
message History {
  oneof request {
    int64 timestamp = 1;
    TimeRange range = 2;
  }
}
```

might clearer on what the timestamp represents - when reading this the first time i thought it was a start without end vs. a stream or snapshot

rpc/gnmi/gnmi-history.md
```diff
@@ -0,0 +1,87 @@
...
+request may return multiple updates per leaf, so the subscription mode for a
+range request must be `STREAM`. If the mode of a range request is not `STREAM`,
+the server should return an `INVALID_ARGUMENT` error. For the paths specified in
+the `SubscriptionList`, the server must transmit all updates it's aware of that
+occurred at the start time or later and before the end time specified in the
+`TimeRange`, and then return. Whether the server should transmit the specified
+data tree as it existed at the start time is determined as usual by the
+`updates_only` field on the `SubscriptionList` message.
+
+If `TimeRange`'s start time is later than its end time, the server must return
```
mhines01, line 71: I feel like this going to be a little more complicated in the expectations from the service providing the stream...

If you have something that both going to serve historical data as well as current data how the data is muxed into those streams is likely to be complicated.

If you want it easy on the client - you would say start at the beginning and just play our the stream from history and then once done start sending all future updates from the live stream but that could have huge buffering problems on the server side of things.

The other option is to allow the streams to be mixed and just play the updates out of order but that would like cause clients problems as they wouldn't know when the "history" stream is actually done or they would just drop all non latest updates as currently for the clients i use - receiving out of order updates -- the old ones are just dropped since they have been replaced by new data anyway

rpc/gnmi/gnmi-history.md
jcostaroberts: I tried to address this below, with "Whether the server should transmit the specified data tree as it existed at the start time is determined as usual by the `updates_only` field on the `SubscriptionList` message."

Maybe a clearer way to say this is "As usual, the server should also transmit the specified data tree as it existed at the subscription start time unless the `updates_only` field on the `SubscriptionList` message is set." What do you think?

rpc/gnmi/gnmi-history.md
jcostaroberts: My reading of the gNMI spec is that updates needn't be buffered until the sync_response message has been sent:

"When the target has transmitted the initial updates for all paths specified within the subscription, a SubscribeResponse message with the sync_response field set to true MUST be transmitted to the client to indicate that the initial transmission of updates has concluded. This provides an indication to the client that all of the existing data for the subscription has been sent at least once. For STREAM subscriptions, such messages are not required for subsequent updates."

So that should cover this case as well.

robshakir: I think that in the scenario  that you have a client that wants to take both "historical and future state", then the behaviour that seems logical (and the lowest implementation overhead) is:

 * play all updates starting at `start_time`
 * once this has completed, send current version of the tree

For a history-consuming client, there is already a need to parse more than a single update to a path within the stream. For example, if we are drawing some graph, then we insert a point at all received datapoints. The buffering would ensure no discontinuities in this graph - but I suspect such a client should rather implement:

- history query for $start_time -> $now
- live subscription

if it is interested in both live and historical data. Or simply a history query if it already knows the time bounds.

jcostaroberts: I'm not sure I understand the first case you're describing--there's no combination of start and end time that should produce updates beginning at a historical start time and then the whole current tree. If updates_only isn't set on a TimeRange subscription, the server should send the tree as it existed at the specified start time, and then incremental updates from then up until the specified end.

In the second case, I agree that clients may just do a historical query + a normal subscription, but are you saying that something about the spec should change, or just pointing out that clients always have the option of implementing open-ended historical queries this way?

robshakir: I'm not sure that the "history + whole current tree" is forbidden in the current gNMI specification, right? I agree that the optimal approach is not to re-send the current tree (since this might be sending duplicates), but there's no specific requirement that a server MUST NOT ever send the same update twice (in-fact, this is required for a solely SAMPLE subscription). I was just pointing out that Marcus' concern on implementation complexity might be mitigated in this manner.

For the latter, I was also intending to support this implementation -- and was just trying to highlight that I think client implementation doesn't necessarily need to change here (for a use case where having not seen a particular set of updates that were dropped/suppressed by the server whilst replaying the history would have a material impact).

jcostaroberts: How would you specify "history + whole current tree"? The only way to get the full tree with a `STREAM` subscription is to set `updates_only` to `false`. If the History extension is used, this will return the tree at the `TimeRange`'s specified `start` time, and then any updates that occurred between that time and the specified `end` time. Thus the only way to get the current tree with the History extension is to specify a `start` time of time.Now, which means no historical data. So history + whole current tree is not possible.

My understanding of Marcus' concern is that it's about whether, in servicing a `STREAM` subscription with `start` < `now` < `end`, the server should buffer updates that occur between `now` and `end` until the initial state has been synchronized (i.e., until the server has sent a `sync_response`), or just transmit them as soon as they occur. That is, I think Marcus' point is about update ordering, not duplication. And by my reading, the spec doesn't prohibit sending current updates before the initial tree has synchronized.

rpc/gnmi/gnmi-history.md
robshakir: This sounds good to me.

rpc/gnmi/gnmi-history.md
robshakir: Apologies, I think we're talking past each other here. I am not saying that we need a way to be able to have the updates plus the whole current tree, I'm saying that if an implementation honoured a history approach by implementing this way then this would not be in violation of the spec.

Regarding the buffering, I agree with you - there is no requirement for ordering in the stream in the current gNMI. I don't see that we should introduce this now.

rpc/gnmi/gnmi-history.md
robshakir: Maybe even `at_timestamp` or `snapshot_timestamp` would be clearer. I agree with you that coming back this doesn't look as clear as it could do.

robshakir:
@jcostaroberts - apologies for the delay here, this got tail-dropped with some personal circumstances that I had. This generally LGTM, if we can make the couple of clarifications above then I think we can merge it.

rpc/gnmi/gnmi-history.md
jcostaroberts: I've changed the field name here (and in gnmi_ext.proto) to `snapshot_time`. I like `at_time` better as a standalone name but making one field a prepositional phrase and the other a noun kind of stresses me out. And went with `time` over `timestamp` because to me "timestamp" indicates a time at which something _happened_, whereas the client here isn't indicating anything happening at all--it just wants the tree at the specified time. Let me know if you have any objection.

rpc/gnmi/gnmi-history.md
```diff
@@ -0,0 +1,87 @@
...
+request field is a `oneof` field whose type indicates which of these request
+types applies.
+
+In all cases, times are represented as int64s interpreted as nanoseconds since
+the Unix Epoch.
+
+### 2.2.1 Snapshot requests
+
+If the `snapshot_time` field is set, the request is for a snapshot. The
+subscription mode for a time request must be `ONCE`; otherwise the server must
```
aashaikh, line 51: maybe clearer to say "... mode for a snapshot request must be ..."

rpc/gnmi/gnmi-history.md
```diff
@@ -0,0 +1,87 @@
...
+return an `INVALID_ARGUMENT` error. The server must transmit the data indicated
+in the `SubscribeRequest` as it existed on the server at the specified time. If
+the specified time indicates a future time at the point the server receives the
+request, the server may either return an `UNIMPLEMENTED` error or choose to wait
+until the specified time has been reached.
+
+### 2.2.2 Stream requests
```
aashaikh, line 58: I'd suggest to call this section "Range requests" or similar ; I think STREAM is just the type of subscription to use for timerange requests.

rpc/gnmi/gnmi-history.md
```diff
@@ -0,0 +1,87 @@
...
+
+If the `range` field is set, the request is for the set of updates that occurred
+during the half-open interval between the specified start and end times. Such a
+request may return multiple updates per leaf, so the subscription mode for a
+range request must be `STREAM`. If the mode of a range request is not `STREAM`,
+the server should return an `INVALID_ARGUMENT` error. For the paths specified in
+the `SubscriptionList`, the server must transmit all updates it's aware of that
+occurred at the start time or later and before the end time specified in the
+`TimeRange`, and then return. As usual, the server should also transmit the
```
aashaikh, line 67: nit: suggest to omit "and then return" or maybe just say "and then close the RPC"

aashaikh (commented):
This LGTM -- just added a couple of minor suggestions.  Leaving for  @robshakir to set approval bit.

rpc/gnmi/gnmi-history.md
jcostaroberts: Thanks--all good catches.

robshakir (approved):
LGTM, thanks for all the changes here.

jcostaroberts:
Thanks! Are we also in a good place on https://github.com/openconfig/gnmi/pull/80 ?

robshakir:
Yes, I merged this internally earlier and will do a push to the gNMI repo
this afternoon.

Thanks,
r.

Issue #139, opened 2021-05-07
JSON format for gNMI Set on a specific list node vs. wildcard list node


wenovus:
### Background
RFC7951 specifies the JSON format for a [list node](https://tools.ietf.org/html/rfc7951#section-5.4), but fails to specify the format for a specific element in a list node (e.g. `/interfaces/interface[name="foo"]`). The gNMI spec also [fails](https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-specification.md#231-json-and-json_ietf
) to provide an example for this case.

### My Questions
1. Is gNMI Set on a specific list node allowed (e.g. `SetRequest.replace.path` is populated with `/interfaces/interface[name="foo"]`)?
1. If so, what should be its JSON format?

aashaikh:
Since the RFC specifies the list encoding to be prefixed by the list name, wouldn't we end up having the data payload to be something like
```json
"interface": [
  {
    "name": "foo",
    "config": {
      "description": "bar"
    }
  }
]
```
If this was a replace operation, it would of course clear the rest of the list element content back to empty/default.

robshakir:
I don't quite agree with @aashaikh here.

If we have a path that refers to an object within a list - which is defined to be an array of objects - then the encoding for a specific entry in a list should be a JSON object. Essentially a list is treated as an indexable array/slice of containers - so using a JSON object means that we're consistent with that interpretation. This seems to be what RESTCONF does from a few examples elsewhere too - so we're consistent there.

So in the OpenConfig interfaces schema, I think we would have:

path: `/interfaces/interface[name=eth0]`
value: 

```json
{
  "name": "eth0",
  "config": {
     "name": "eth0",
     "description": "ethernet zero"
  }
}
```

robshakir:
This also seems more logical than having a `replace` to something more specific than the `list` node clearing what appear to be unrelated entries.

For example, replace `/interfaces/interface[name=eth0]` replacing `/interfaces/interface[name=eth1]` would be quite unexpected to me.

aashaikh:
We should definitely do what is consistent with well-understood behavior here -- so the object rather than array SGTM.

But it seemed a little odd that the list is encoded as an array, but a single element in the list would be an object rather than a single item array containing the object.

robshakir:
I think this is consistent with what you would expect in most languages though, right? For example if I write:

```c
int n[5];

n = {0,1,2,3,4}; // refer to the array as an array
[...]

PR #140, opened 2021-05-10
Proposal for clarifications when calling gNMI Set on a list node or element.

wenovus:
1. Add example JSON_IETF format for updating an individual list element.
1. Clarify that an empty subset of attributes for a keyed-list is unacceptable as a path for gNMI Set.
1. Clarify that in OpenConfig, MUST specify a path that terminates on the enclosing container in order to modify a whole list node.
1. Minor grammar fixes.

### Background
I found these points to be missing from either this doc or RFC7951 (#139). After some discussions with Anees, I'm proposing to add these clarifications that may or may not be the standard or established convention. In that case, modification suggestions are welcome to correct any inaccuracies.

### DIFF
--- a/rpc/gnmi/gnmi-specification.md
+++ b/rpc/gnmi/gnmi-specification.md
@@ -470,13 +470,12 @@ update: <
     >
   >
   val: <
-    json_ietf_val: `{ "b": {
-                        "name": "b1",
-                        "c": {
-                          "d": "AStringValue",
-                          "e": 10042
-                        }
-                   }
+    json_ietf_val: `{ 
+                      "name": "b1",
+                      "c": {
+                        "d": "AStringValue",
+                        "e": 10042
+                      }
             }`
   >
 >
@@ -1138,7 +1137,7 @@ array), the following considerations apply:
 
 *   In the case that multiple attribute values are required to uniquely address
     an element - e.g., `/a/f[k1=10][k2=20] `- and a replace or update
-    operation's path specifies a strict subset of the attributes (e.g.,
+    operation's path specifies a proper subset of the attributes (e.g.,
     `/a/f[k1=10]`, `/a/f`), then this MUST be considered an error by the target
     system - and an status code of` InvalidArgument (3)` specified.
 *   Where the path specified refers to a node which itself represents the
@@ -1150,11 +1149,6 @@ array), the following considerations apply:
     as their own elements within the data tree, update or replace operations
     that modify instances of the key in conflicting ways MUST be considered an
     error. The target MUST return a status code of `InvalidArgument (3)`.
-*   In OpenConfig, a path which refers to a keyed YANG list node for
-    modification (as opposed to an element of the list node) MUST terminate at
-    the [enclosing
-    container](https://github.com/openconfig/public/blob/master/doc/openconfig_style_guide.md#list)
-    element.
 
 For example, consider a tree corresponding to the examples above, as illustrated
 below.


rpc/gnmi/gnmi-specification.md
```diff
@@ -1109,18 +1138,23 @@ array), the following considerations apply:
 
 *   In the case that multiple attribute values are required to uniquely address
     an element - e.g., `/a/f[k1=10][k2=20] `- and a replace or update
-    operation's path specifies  a subset of the attributes (e.g., `/a/f[k1=10]`)
-    then this MUST be considered an error by the target system - and an status
-    code of` InvalidArgument (3)` specified.
+    operation's path specifies a strict subset of the attributes (e.g.,
+    `/a/f[k1=10]`, `/a/f`), then this MUST be considered an error by the target
```
robshakir, line 1142: This change means that one cannot send `/path/to/list` with a body that is an array of objects in JSON (assuming JSON_IETF) right? Is there a reason to disallow this?

rpc/gnmi/gnmi-specification.md
```diff
@@ -453,6 +453,35 @@ update: <
...
+  >
+  val: <
+    json_ietf_val: `{ "b": {
+                        "name": "b1",
+                        "c": {
+                          "d": "AStringValue",
+                          "e": 10042
+                        }
+                   }
+            }`
```
wenovus, line 480: Is this correct? should the "b"'s level be taken out?

rpc/gnmi/gnmi-specification.md
wenovus: I didn't change anything here -- a strict/proper subset is likely what the original sentence meant to say.

I think there is no clear reason to disallow this, since the keys for each element in the array should be provided in the update message as well. However, that would now mean this spec needs to be changed, right?

rpc/gnmi/gnmi-specification.md
wenovus: Yes, I'll fix it: https://github.com/openconfig/reference/issues/139#issuecomment-840829882

rpc/gnmi/gnmi-specification.md
```diff
@@ -1109,18 +1138,23 @@ array), the following considerations apply:
...
-    collection of objects (list, map, or array) a replace operation MUST remove
+    collection of objects (list, map, or array), a replace operation MUST remove
     all collection entries that are not supplied in the value provided in the
     `SetRequest`. An update operation MUST be considered to add new entries to
     the collection if they do not exist.
 *   In the case that key values are specified both as attributes of a node, and
     as their own elements within the data tree, update or replace operations
     that modify instances of the key in conflicting ways MUST be considered an
     error. The target MUST return a status code of `InvalidArgument (3)`.
+*   In OpenConfig, a path which refers to a keyed YANG list node for
```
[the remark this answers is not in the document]
wenovus: The above question refers to modifying a specific list element, whereas this question refers to modifying a whole list, so I think they're unrelated from one another.

rpc/gnmi/gnmi-specification.md
```diff
@@ -453,6 +453,34 @@ update: <
...
+                      "name": "b1",
+                      "c": {
+                        "d": "AStringValue",
+                        "e": 10042
+                      }
+            }`
+  >
+>
+```
+
```
robshakir, line 483: Is it worth explicitly stating here that the contents of the update is the JSON object that would be expected at that list entry?

rpc/gnmi/gnmi-specification.md
robshakir: I think there's an ambiguity here, I think that this was intended to say a non-null subset - i.e., keys were partially specified.

I think there are a few cases here:

1. No keys are specified - e.g., `/interfaces/interface` - in this case, I think I expect that the serialised content of the list would be specified. e.g., the array of objects for JSON_IETF.
2. A partial set of keys are specified - this is an error with `InvalidArgument` in my view.

I think it'd be worth checking implementations to see whether the no-keys variant is supported. This was the original motivation for having surrounding containers in OpenConfig such that a get + set could be done on the surrounding container to get all entries.

robshakir (changes requested):

Issue #142, opened 2021-07-28
What should a gNMI server return when a default value is queried?


wenovus:
There are four independent cases considered below. All examples use the /interfaces/interface/state/enabled leaf, which has a default value of `true`.

## Question
For all queries below, consider the case that the interface at /interfaces/interface[name="foo"] exists, and that its "enabled" field has its default value of `true`.

The question for all of them is, must ("/interfaces/interface[name="foo"]/state/enabled", `true`) be returned by GET?

### 1 Non-wildcard leaf query
e.g. GET /interfaces/interface[name="foo"]/state/enabled

### 2 Non-wildcard non-leaf query
e.g. GET /interfaces/interface[name="foo"]

### 3 Wildcard leaf query
e.g. GET /interfaces/interface[name="*"]/state/enabled

### 4 Wildcard non-leaf query
e.g. GET /interfaces/interface[name="*"]

wenovus:
@dan-lepage pointed out the interesting case where `/interfaces/interface[name="foo"]` does not exist. According to the [spec](https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-specification.md#3513-the-subscription-message), GET should also return nothing since it's a valid path. So, if GET on `/interfaces/interface[name="foo"]/state/enabled` doesn't return the default value, then it would be impossible to distinguish between a value existing as its default value, and a value that doesn't exist due to one of its an ancestor's list element not existing.

By requiring GET to always return a value for a leaf GET/SUBSCRIBE, even if it's the default value, it removes this ambiguity.

This is also relevant for non-leafs. While I'm guessing that list keys should not have default values specified, it's possible to have containers underneath lists that make it possible for the same ambiguity to occur. Although this case may occur less often, we may still need to consider it for completeness.

Related question: Are we planning on supporting presence-containers?

dan-lepage:
> According to the [spec](https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-specification.md#3513-the-subscription-message), GET should also return nothing since it's a valid path.

If I'm reading the spec correctly, it doesn't actually say what should happen on a GET. On a SUBSCRIBE with ONCE set, it should actually never return - "In the case that a particular path does not (yet) exist, the target MUST NOT close the RPC, and instead should continue to monitor for the existence of the path, and transmit telemetry updates should it exist in the future."

[...]

PR #144, opened 2021-09-10
Clarify wording on SUBSCRIBE/GET to a non-existent path.

wenovus:
See discussion at https://github.com/openconfig/reference/issues/142#issuecomment-917086760

### DIFF
--- a/rpc/gnmi/gnmi-specification.md
+++ b/rpc/gnmi/gnmi-specification.md
@@ -917,7 +917,9 @@ The `GetResponse` message consists of:
     and hence MUST NOT collapse data from multiple paths into a single
     `Notification` within the response. The `timestamp` field of the
     `Notification` message MUST be set to the time at which the target's
-    snapshot of the relevant path was taken.
+    snapshot of the relevant path was taken. If the path is syntactically valid
+    but does not (yet) exist, then the `update` field of the `Notification` MUST
+    be empty.
 *   `extension` - a repeated field used to carry gNMI extensions, as per the
     description in [Section 2.7](#27-extensions-to-gnmi).
 
@@ -1400,8 +1402,10 @@ the current data tree on the server. While the path within the subscription
 SHOULD be a valid path within the set of schema modules that the target
 supports, subscribing to any syntactically valid path within such modules MUST
 be allowed. In the case that a particular path does not (yet) exist, the target
-MUST NOT close the RPC, and instead should continue to monitor for the existence
-of the path, and transmit telemetry updates should it exist in the future.
+would send just a `sync_response` for `ONCE` and `POLL` subscriptions, but MUST
+NOT close the RPC for `STREAM` subscriptions, and instead should continue to
+monitor for the existence of the path, and transmit telemetry updates should it
+exist in the future.
 
 <!-- TODO(robjs): Need a way to be able to send this information. Previously
 said:


rpc/gnmi/gnmi-specification.md
```diff
@@ -1400,8 +1402,10 @@ the current data tree on the server. While the path within the subscription
 SHOULD be a valid path within the set of schema modules that the target
 supports, subscribing to any syntactically valid path within such modules MUST
 be allowed. In the case that a particular path does not (yet) exist, the target
-MUST NOT close the RPC, and instead should continue to monitor for the existence
-of the path, and transmit telemetry updates should it exist in the future.
+would send just a `sync_response` for `ONCE` and `POLL` subscriptions, but MUST
```
dan-lepage, line 1405: I think the word "would" ought to be MUST, though I also think this might be clearer to make two different statements for the two different cases, and make it clearer how this applies when you're subscribing to multiple paths, some of which exist and some of which don't. For example, maybe:

> When a `GET` or `POLL` subscription includes paths that do not (yet) exist in the data tree, the target MUST NOT include those paths in its response (for example, a `GET` for exclusively paths that don't exist should return a `SubscriptionResponse` that does not contain any updates).

> When a `STREAM` subscription includes paths that do not (yet) exist in the data tree, the target MUST continue to monitor for the existence of those paths, and transmit telemetry updates should any of them exist in the future. In particular, a `STREAM` subscription consisting _solely_ of paths that are not in the current data tree MUST NOT close the RPC, but instead should continue monitoring in case those paths are created later on.

dan-lepage: One question: By default, a STREAM starts by sending a notification of the current state; I assume in the case that none of the subscribed paths are set it should still send this notification, it just won't contain any updates?

wenovus: I'm happy with your modification, although I assume by `GET` you mean `ONCE`?

> One question: By default, a STREAM starts by sending a notification of the current state; I assume in the case that none of the subscribed paths are set it should still send this notification, it just won't contain any updates?

This is also my understanding from https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-specification.md#3523-sending-telemetry-updates

dan-lepage: Oh sorry, yes, I meant ONCE.

dplore: @gcsl Can you give us your review?

rpc/gnmi/gnmi-specification.md
```diff
@@ -917,7 +917,9 @@ The `GetResponse` message consists of:
     and hence MUST NOT collapse data from multiple paths into a single
     `Notification` within the response. The `timestamp` field of the
     `Notification` message MUST be set to the time at which the target's
-    snapshot of the relevant path was taken.
+    snapshot of the relevant path was taken. If the path is syntactically valid
+    but does not (yet) exist, then the `update` field of the `Notification` MUST
```
robshakir, line 921: I think here the target should respond with an error using the `NotFound` status code. This seems (to me) to be the most explicit thing to do when a client is explicitly asking for a very particular path.

rpc/gnmi/gnmi-specification.md
```diff
@@ -1400,8 +1402,10 @@ the current data tree on the server. While the path within the subscription
 SHOULD be a valid path within the set of schema modules that the target
 supports, subscribing to any syntactically valid path within such modules MUST
 be allowed. In the case that a particular path does not (yet) exist, the target
-MUST NOT close the RPC, and instead should continue to monitor for the existence
-of the path, and transmit telemetry updates should it exist in the future.
+would send a `sync_response` as the very first message to indicate this, and in
```
robshakir, line 1405: I still don't like this wording.

I suggest:

```
In the case that a particular path does not (yet) exist, the target should:
 * For `ONCE` subscriptions, return a `sync_response` and close the RPC.
 * For `POLL` subscriptions, return a `sync_response` immediately, and await the next polling request from the client.
 * For `STREAM` subscriptions, if there is no data at a particular path at the start of a subscription, or when a path is sampled, the target should return a `sync_response` immediately. Should the path be created at some later time, the target should begin to transmit updates for the path, according to the specified mode of the subscription.
```

Thoughts?

robshakir (changes requested):

rpc/gnmi/gnmi-specification.md
wenovus: Wouldn't this deviate from Subscribe where no error is produced?

> There is no requirement that the path specified in the message must exist within the current data tree on the server. While the path within the subscription SHOULD be a valid path within the set of schema modules that the target supports, subscribing to any syntactically valid path within such modules MUST be allowed. In the case that a particular path does not (yet) exist, the target MUST NOT close the RPC, and instead should continue to monitor for the existence of the path, and transmit telemetry updates should it exist in the future.

rpc/gnmi/gnmi-specification.md
wenovus: I like it. One question on a previous sentence:

> subscribing to any syntactically valid path within such modules

So, if the path is not schema-compliant, it should return `NOT_FOUND`? i.e. https://pkg.go.dev/google.golang.org/grpc/codes#Code

Should we document this as well?

PR #146, opened 2021-10-19
Clarify that SubscribeResponse messages are not guaranteed to be in timestamp order.

### LINKED ISSUE #145: Must STREAM Subscriptions' responses be in timestamp order?
https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-specification.md#35152-stream-subscriptions

Sorry I remember this not being guaranteed. Could the spec clarify this?

wenovus:
I put this as another paragraph under "The SubscribeResponse Message". Let me know if you'd like this to be in a separate section.

FIXED #145

### DIFF
--- a/rpc/gnmi/gnmi-specification.md
+++ b/rpc/gnmi/gnmi-specification.md
@@ -1434,6 +1434,11 @@ established `Subscribe` RPC. The message contains the following fields:
 *   `extension` - a repeated field used to carry gNMI extensions, as per the
     description in [Section 2.7](#27-extensions-to-gnmi).
 
+`SubscribeResponse` messages with the `update` field set to a `Notification`
+message are not guaranteed to be sent in timestamp order. This allows congestion
+management schemes to provide the most recent data while avoiding starving any
+particular path.
+
 #### 3.5.1.5 Creating Subscriptions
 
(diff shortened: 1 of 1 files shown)


rpc/gnmi/gnmi-specification.md
```diff
@@ -1434,6 +1434,11 @@ established `Subscribe` RPC. The message contains the following fields:
 *   `extension` - a repeated field used to carry gNMI extensions, as per the
     description in [Section 2.7](#27-extensions-to-gnmi).
 
+`SubscribeResponse` messages with the `update` field set to a `Notification`
+message are not guaranteed to be sent in timestamp order. This allows congestion
```
robshakir, line 1438: I would just leave this at the first sentence, and maybe say "are not required to be sent, and must not be assumed to be in timestamp order".

wenovus: Do you mean that a message set with `Notification` is not required, or that the timestamp itself is not required?

wenovus: Ok, PTAL, I changed to "MUST NOT assume that these **messages** will be received", since there can only be a single timestamp in each message.

robshakir (approved):
LGTM, thanks for updating the spec :-)

PR #148, opened 2021-11-16
Remove client and server side aliases from gNMI

gcsl:
Aliases were initially added to the specification in anticipation of tuning bandwidth of transmission for streaming Subscribe RPCs. To our knowledge these have not been implemented by any vendor or network operator to date.  I am proposing the removal to clean up the specification to be closer to existing implementations and avoid confusion for new implementations.


marcushines:
I don't have an objection at this point.
As you stated i don't see anyone clamoring for this feature and it is only removes complexity. go for it

PR #149, opened 2021-11-16
Remove heartbeat from gNMI

gcsl:
I am proposing to remove the heartbeat from the gNMI specification as it is redundant implementation to that which can be achieved otherwise.  E.g. a client could mix ON_CHANGE and SAMPLE subscriptions to get the same effect.


gwizdms:
Several of our vendors currently use this feature with dial-in or have plans to develop the feature in releases for 2022.  Although, creating two sensors (e.g. ON_CHANGE and SAMPLE) could net the same effect, a simpler operational management solution is to have a definition of the feature "heartbeat" that is not forgotten with two subscriptions.

gcsl:
My intention with this proposal was to remove a supposed unused part of the specification.  Since you have active and anticipated uses and multiple vendors participating, I will close this issue.

PR #150, opened 2021-11-16
Create decimal64-deprecation.md

gcsl:
Markdown version of the document presented to the Open Config Operators Group meeting on April 6, 2021.

### DIFF
--- a/rpc/gnmi/decimal64-deprecation.md
+++ b/rpc/gnmi/decimal64-deprecation.md
@@ -43,7 +43,7 @@ floating point for real numbers.  All major programming languages have native
 support for binary floating point as do the vast majority of modern CPU hardware
 where gNMI servers or its clients would likely run.
 
-Decimal64, on the other hand, is “ intended for applications where it is
+Decimal64, on the other hand, is “intended for applications where it is
 necessary to emulate decimal rounding exactly, such as financial and tax computations.”
 ([Wikipedia](https://en.wikipedia.org/wiki/Decimal64_floating-point_format)) The
 decimal encoding is used for financial transactions because binary fractions
--- a/rpc/gnmi/decimal64-deprecation.md
+++ b/rpc/gnmi/decimal64-deprecation.md
@@ -5,7 +5,7 @@
 
 ### Decision
 
-Deprecate and schedule removal of the Decimal64 message from gNMI and stipulate
+This document details the proposal (and decision, once approved) to deprecate and schedule removal of the Decimal64 message from gNMI and stipulate
 all Decimal64 values in the OpenConfig models be delivered as
 gNMI.TypedValue.double_val.
 
--- a/rpc/gnmi/decimal64-deprecation.md
+++ b/rpc/gnmi/decimal64-deprecation.md
@@ -3,7 +3,7 @@
 **Updated**: April 28, 2022
 **Author**: Carl Lebsack
 
-### Proposal
+### Decision
 
 Deprecate and schedule removal of the Decimal64 message from gNMI and stipulate
 all Decimal64 values in the OpenConfig models be delivered as
--- a/rpc/gnmi/decimal64-deprecation.md
+++ b/rpc/gnmi/decimal64-deprecation.md
@@ -163,6 +163,6 @@ organizational human communication. The gNMI specification and Protocol Buffer
 definition include two ways of rendering fractional values, Decimal64 and
 double_val.  Because of this, we have some vendors using one or the other or both
 methods for various paths.  Downstream clients need to handle both cases and
-convert Decimal64 to float where it appears.  There is often the question of,
+convert Decimal64 to double where it appears.  There is often the question of,
 “What matches the official OpenConfig schema”.  The goal of this proposal is to
 codify that answer as double_val for gNMI.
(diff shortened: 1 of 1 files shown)


gcsl:
It has been brought to my attention that there are some use cases of gNMI with non-OpenConfig models where the underlying data is modeled as Decimal64 but cannot be represented with the precision of single precision floating point. I am told that these known uses all use double precision within the internal implementations, and not a native Decimal64 representation.

In keeping with simplicity of gNMI using a small number of types, I think the most likely way forward here is to add a double precision float to TypedValue and deprecate the single precision (whilst maintaining both for some period of time).  Despite double precision being overkill for most OC modeled data today, I am aware that many clients are likely converting to double precision anyway and have been told by some vendors that this is the type chosen within internal implementations in many cases also.

rpc/gnmi/decimal64-deprecation.md
```diff
@@ -0,0 +1,168 @@
...
+
+### Why is Decimal64 problematic for gNMI?
+
+The primary problem is inefficiency which manifests in multiple ways by virtue
+of a misalignment of the natural encoding of the real data and the decimal
+encoding of the Decimal64 type. Before enumerating some of the inefficiencies in
+practical use, we’ll start with a brief discussion of the ultimate source of the
+data being modeled and the inappropriateness of Decimal64 for encoding.
+
+All values within the OpenConfig models using the YANG Decimal64 type are
```
robshakir, line 30: One note here is that values that are /actually/ floating point values (e.g., in IS-IS we have fields that are float32 because they are received in LSPs as float32) use a specific `ieeefloat32` `typedef` already.

rpc/gnmi/decimal64-deprecation.md
```diff
@@ -0,0 +1,168 @@
...
+fundamentally binary because they originate from
+[Analog to Digital Converters](https://en.wikipedia.org/wiki/Analog-to-digital_converter)
+(ADC) whereby a continuous analog signal, such as the laser power in an optical
+transceiver, is discretized into a digital binary representation. This means
+that despite the fact that we may render the outputs in a decimal form when
+presenting a human readable output, the data itself is binary.
```
robshakir, line 36:
```suggestion
presenting a human readable output, the data itself is binary. It would likely have been preferable to use a native floating point typedef (such as the `ieeefloat32` typedef that was defined within `openconfig-types`) as opposed to `decimal64`. The continued use of `decimal64` is primarily for backwards-compatibility reasons.
```

rpc/gnmi/decimal64-deprecation.md
```diff
@@ -0,0 +1,168 @@
...
+
+Further, it is expected that a majority of clients using this data will
+ultimately be using binary floating point representation and Decimal64 is simply
+an awkward go-between induced by virtue of the YANG decision on only supporting
+a single encoding for real types.  All time-series databases examined use binary
+floating point for real numbers.  All major programming languages have native
+support for binary floating point as do the vast majority of modern CPU hardware
+where gNMI servers or its clients would likely run.
+
+Decimal64, on the other hand, is “ intended for applications where it is
```
robshakir, line 46:
```suggestion
Decimal64, on the other hand, is “intended for applications where it is
```

robshakir (changes requested):
Some changes to make.

I think it would be good to ask folks working on other implementations to just review this document - have you had conversations with these folks directly that are not reflected in this PR?

rpc/gnmi/decimal64-deprecation.md
```diff
@@ -0,0 +1,168 @@
...
+#### Confusing both implementers and operators
+
+Another significant inefficiency of the use of Decimal64 is one of inter-
+organizational human communication. The gNMI specification and Protocol Buffer
+definition include two ways of rendering fractional values, Decimal64 and
+double_val.  Because of this, we have some vendors using one or the other or both
+methods for various paths.  Downstream clients need to handle both cases and
+convert Decimal64 to float where it appears.  There is often the question of,
+“What matches the official OpenConfig schema”.  The goal of this proposal is to
+codify that answer as double_val for gNMI.
```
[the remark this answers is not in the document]
robshakir: ACK, I suggest that at this point this isn't really a proposal any more, since we're basically saying that we're doing it. Will make suggestions for how to reflect that.

rpc/gnmi/decimal64-deprecation.md
```diff
@@ -0,0 +1,168 @@
+## Proposal to deprecate Decimal64 in gNMI
```
robshakir, line 1:
```suggestion
## Deprecation of Decimal64 in gNMI
```

rpc/gnmi/decimal64-deprecation.md
```diff
@@ -0,0 +1,168 @@
...
+
+**Updated**: April 28, 2022
+**Author**: Carl Lebsack
+
+### Proposal
```
robshakir, line 6:
```suggestion
### Decision
```

rpc/gnmi/decimal64-deprecation.md
```diff
@@ -0,0 +1,168 @@
...
+
+Deprecate and schedule removal of the Decimal64 message from gNMI and stipulate
```
robshakir, line 8:
```suggestion
This document details the proposal (and decision, once approved) to deprecate and schedule removal of the Decimal64 message from gNMI and stipulate
```

robshakir (approved):
Suggestions just to make the decision that is being made here clearer, but I am good to merge this.

Thanks for confirming the discussions with other interested parties here.

Issue #155, opened 2022-02-17
Clarifying gNMI server behavior for GET and default values


dplore:
This is a fork of issues from  #142 

* GET for a state path leaf MUST return one of:
    * a value (the default value or a set value)
    * path not found if the path is syntactically valid but is not currently instantiated on the target
    * not supported if the path is syntactically valid but is not implemented
    * an error if the path is syntactically incorrect 

* GET for a config path leaf MUST return one of:
    * the set value OR the default value if defined OR nil if the value is not set and there is no default
    * path not found if the path is syntactically valid but is not currently instantiated on the target
    * not supported if the path is syntactically valid but is not implemented
    * an error if the path is syntactically incorrect 

* GET a subtree
   *  TBD

This may need to be clarified as there is an implementation which is returning 'nil' when a leaf that does exist and contains a default value.

wenovus:
For state paths, could it also return nil (i.e. no updates) if the path is not populated? In other words, should there be a difference between config and state paths as far as telemetry is concerned?

dan-lepage:
The state paths represent the operational state of the device, and the YANG spec requires:

>   When the default value is in use, the server MUST operationally
>   behave as if the leaf was present in the data tree with the default
>   value as its value.
(from [rfc7950: Section 7.6.1](https://datatracker.ietf.org/doc/html/rfc7950#section-7.6.1))

So if a path has a default value and has not been explicitly set, the server MUST return the default value when you query the state.

There are two ways to unset a path - you can REPLACE an ancestor without providing a value for this path, or you can DELETE the path directly. In the first case, [the gNMI spec](https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-specification.md#344-modes-of-update-replace-versus-update) says that the value shouldn't become unset if it has a default, but should instead be set to its default value [Section 3.4.6](https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-specification.md#346-deleting-configuration) says the opposite for deletion - if a path is deleted, it should be removed from the data tree entirely. I think this is an error, and section 3.4.6 should instead say that DELETE behaves just like an empty REPLACE, and sets a path to its default value if there is one.

This means that we won't need to handle config and state paths differently, and it means that the implementation of GET doesn't even need to know about defaults - it's impossible for a node that has a default value to ever be unset.

[...]

PR #157, opened 2022-02-22
Update gNMI specification to clarify `Set` error handling.

### LINKED ISSUE #62: errors in SetResponse (proto vs. spec)
Looking at the latest gnmi.proto, I see that Error has been deprecated, as has all its uses, including within SetResponse and UpdateResult.

However Section 3.4 still says that "In the case of a failure of an operation, the status of the UpdateResult message MUST be populated with error information as per the specification in Section 3.4.7.".

Similarly, 3.4.7 still says that we "MUST set the message field of the UpdateResult corresponding to the failed operation to an Error message indicating failure. In the case that the processed operation is not the only operation within the SetRequest the target MUST set the message field of the UpdateResult messages for all other operations, setting the code field to Aborted (10).".

I woulds image the proto is correct and the spec just hasn't been updated to agree yet - can this be clarified?

robshakir:
```
  * (M) rpc/gnmi/gnmi-specification.md
    - remove references to the deprecated `error` field in the
      `SetResponse` and `UpdateResult` messages, explicitly
      specifying the error handling behaviour follows the standard
      gRPC error model.
```

Fixes #62.

### DIFF
--- a/rpc/gnmi/gnmi-specification.md
+++ b/rpc/gnmi/gnmi-specification.md
@@ -4,10 +4,10 @@
 Paul Borman, Marcus Hines, Carl Lebsack, Chris Morrow, Anees Shaikh, Rob Shakir
 
 **Date:**
-January 30, 2018
+February 22, 2022
 
 **Version:**
-0.6.0
+0.8.0
 
 # Table of Contents
 
@@ -963,13 +963,9 @@ message. For each operation specified in the `SetRequest` message, an
 `UpdateResult` message MUST be included in the response field of the
 `SetResponse`. The order in which the operations are applied MUST be maintained
 such that `UpdateResult` messages can be correlated to the `SetRequest`
-operations. In the case of a failure of an operation, the status of the
-`UpdateResult` message MUST be populated with error information as per the
-specification in [Section 3.4.7](#347-error-handling). In addition, the status
-of the `SetResponse` message MUST be populated with an error message indicating
-the success or failure of the set of operations within the `SetRequest` message
-(again using the error handling behavior defined in [Section
-3.4.7](#347-error-handling)).
+operations. In the case of any update failing to be applied, the error
+handling behaviours described in [Section 3.4.7](#347-error-handling) must be
+followed.
 
 ### 3.4.1 The SetRequest Message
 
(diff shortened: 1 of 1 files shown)


wenovus (approved):
LGTM

wenovus (commented):
@dplore noted here that there is ambiguity as to what to return when multiple error codes are valid: https://github.com/openconfig/reference/issues/155#issuecomment-1179134075

dplore:
In the case of multiple errors, I believe we should specify "the implementation SHOULD return the first error encountered"

As a newcomer to gRPC, I found it useful to directly reference the  [standard gRPC return codes](https://github.com/googleapis/googleapis/blob/4d8b76dbfd1ce8d7b1fcd13df2d2a39e7ccc2c9f/google/rpc/code.proto#L162) and this [gRPC core doc page on status codes](https://grpc.github.io/grpc/core/md_doc_statuscodes.html)

rpc/gnmi/gnmi-specification.md
```diff
@@ -1207,9 +1191,9 @@ canonical gRPC error codes:
     invalid value within the `Update` message specified - `InvalidArgument (3)`.
     This error SHOULD be used in cases where the payload specifies scalar values
     that do not correspond to the correct schema type.
-* When the client specifies a payload utilising an encoding that is not
-  supported by the target (e.g., JSON) - `Unimplemented (12)` SHOULD be used to
-  indicate that the encoding is not supported.
+*   When the client specifies a payload utilising an encoding that is not
+    supported by the target (e.g., JSON) - `Unimplemented (12)` SHOULD be used to
+    indicate that the encoding is not supported.
```
dplore, line 1197:
```suggestion
*  When the client specifies a value which is not permitted by the target - `Unimplemented  (12)`
    SHOULD be used.  For example, setting an interface full_duplex leaf to false when the interface
    only supports full_duplex. 
```

dplore (commented):
Please consider comments on 
- additional use case for unimplemented (12) and In the case of multiple errors
- I believe we should specify "the implementation SHOULD return the first error encountered"

As a newcomer to gRPC, I found it useful to directly reference the [standard gRPC return codes](https://github.com/googleapis/googleapis/blob/4d8b76dbfd1ce8d7b1fcd13df2d2a39e7ccc2c9f/google/rpc/code.proto#L162) and this [gRPC core doc page on status codes](https://grpc.github.io/grpc/core/md_doc_statuscodes.html)

I found these more straightforward to read than the referenced gRPC error handling guide.

PR #162, opened 2022-03-08
clarification for allow_aggregation control

osquitun:
clarification for allow_aggregation control

### DIFF
--- a/rpc/gnmi/gnmi-specification.md
+++ b/rpc/gnmi/gnmi-specification.md
@@ -347,6 +347,8 @@ For any JSON encoding:
 *   Where the data item referred to has child nodes, the `val` field contains a
     serialised JSON entity (object or array) corresponding to the referenced
     item.
+*   The serialization MUST follow the methods and fields in the 
+    Subscribe RPC in [Section 3.5.1](#351-managing-subscriptions).
 
 Using the following example data tree:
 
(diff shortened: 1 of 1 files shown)


rpc/gnmi/gnmi-specification.md
```diff
@@ -1603,10 +1605,10 @@ affecting the sample accuracy and freshness to the client, and as a result, on
 the client's ability to react to events on the target.
 
 Since it is not possible for the target to infer whether its clients are
-sensitive to the latency introduced by bundling, if a target implements
-optimizations such that multiple `Update` messages are bundled together,
-it MUST provide an ability to disable this functionality within the
-configuration of the gNMI service. Additionally, a target SHOULD provide means
+sensitive to the latency introduced by bundling, the target MUST follow the
+`SubscriptionList` fields in [Section 3.5.1.2](#3512-the-subscriptionlist-message)
+that indicate if `allow_aggregation` is permitted.
```
robshakir, line 1610: I think that maybe we need a larger change to the wording here. I think we should clarify what _bundling_ means here vs. _aggregation_. Specifically, we were trying to give guidance here as to how many `update` messages there were per `notification`.

In aggregation, I think we are looking at how many leaves per `update` there are. i.e., without `allow_aggregation` this should be 1. 

I'd propose we make this clarification, and then add some text as to when aggregation is triggered. WDYT?

gwizdms:
When allow aggregation is set in the SubscriptionList how does the model (schema) know where to aggregate?  An explanation on how it is referenced in the models would help in the specification.

PR #168, opened 2022-07-06
update heading indexes after path prefixes removal

hellt:
In https://github.com/openconfig/reference/pull/148 we removed path prefixes, this PR updates the heading indexes to account for this change


dplore (approved):
LGTM

PR #170, opened 2022-07-06
Subscribe: A transition to a default value must use `update` rather than `delete`

wenovus:
Addresses this question https://github.com/openconfig/reference/issues/142#issuecomment-1119957888 by extracting from the several comments following it.


dplore (approved):

robshakir (approved):

robshakir:
Should we bump the version of the spec here to something like 0.8.1?

wenovus:
> Should we bump the version of the spec here to something like 0.8.1?

Done.

PR #174, opened 2022-11-10
Add special case for origin CLI

dplore:
Clarify use of origin CLI when mixed with other origins.

Resolves https://github.com/openconfig/featureprofiles/issues/740

### DIFF
--- a/rpc/gnmi/mixed-schema.md
+++ b/rpc/gnmi/mixed-schema.md
@@ -113,8 +113,8 @@ Special considerations for mixing native schema are currently undefined.
   the requirements for `origin` `cli` to take precedence.
 * The `replace` operation for a setRequest containing only `origin` `cli` MUST
   be treated as a full device configuration replacement.
-* The `replace` operation for a setRequest containing both CLI and OpenConfig,
-  the OpenConfig and CLI SHOULD be supported.  The content of the two origins
+* The `replace` operation for a setRequest containing both CLI and OpenConfig
+  SHOULD be supported.  The content of the two origins
   SHOULD be be merged, with precedence to CLI for any overlapping
   configuration, and then the replace operation performed.  Since the CLI is
   interpreted as a full configuration replacement, the result of this
--- a/rpc/gnmi/mixed-schema.md
+++ b/rpc/gnmi/mixed-schema.md
@@ -115,7 +115,7 @@ Special considerations for mixing native schema are currently undefined.
   be treated as a full device configuration replacement.
 * The `replace` operation for a setRequest containing both CLI and OpenConfig
   SHOULD be supported.  The content of the two origins
-  SHOULD be be merged, with precedence to CLI for any overlapping
+  SHOULD be merged, with precedence to CLI for any overlapping
   configuration, and then the replace operation performed.  Since the CLI is
   interpreted as a full configuration replacement, the result of this
   setRequest should be a full configuration replacement with the merged
--- a/rpc/gnmi/mixed-schema.md
+++ b/rpc/gnmi/mixed-schema.md
@@ -120,7 +120,7 @@ Special considerations for mixing native schema are currently undefined.
   interpreted as a full configuration replacement, the result of this
   setRequest should be a full configuration replacement with the merged
   content of the CLI and OpenConfig paths.
-* A `replace` operation for `origin` `openconfig`, followed by a `update`
+* A `replace` operation for `origin` `openconfig`, followed by an `update`
   operation for `origin` `cli` in a single setRequest SHOULD be supported.
   The goal is to allow configuration replacement to occur at a sub-tree level
   using both OpenConfig and the CLI configuration data.
--- a/rpc/gnmi/mixed-schema.md
+++ b/rpc/gnmi/mixed-schema.md
@@ -98,7 +98,7 @@ constrained to the specified origin. Particularly:
 
 ### Special considerations for the use of Origin CLI
 
-The following rules apply to a setRequest containing `origin` CLI.  TODO:
+The following rules apply to a `SetRequest` containing `origin` CLI.  TODO:
 Special considerations for mixing native schema are currently undefined.
 
 * If a `SetRequest` contains `origin` `cli` and `origin` `openconfig`, the CLI
--- a/rpc/gnmi/mixed-schema.md
+++ b/rpc/gnmi/mixed-schema.md
@@ -99,6 +99,24 @@ constrained to the specified origin. Particularly:
   the specified `origin`. To delete contents from multiple origins, a client
   MUST specify multiple paths within the `delete` of the `SetRequest`.
 
+### Special considerations for the use of Origin CLI
+* If a `Set` RPC contains `origin` CLI, the CLI origin MUST be given precedence
+  over the OpenConfig origin.  This precedence is independent of the order in
+  which origin CLI appears.  This overrides the gNMI specification requirement
+  of ordered path based precedence within a `SetRequest`.  The goal of this
+  precedence is to create deterministic behavior for resolving overlapping
+  configuration.
+* The `update` operation MUST be supported for `origin` CLI.  The `update`
+  operation is compatible with the requirement for `origin` CLI to take
+  precedence.
+  
+* Because there is no explicit expectation of CLI to use a path based
+  structure, use of the `replace` operation is not recommended unless the
+  intent is to replace the entire configuration with the contents of the origin CLI message.  Replacing
+  the entire configuration with CLI content in a mixed origin `SetRequest`
+  where origin CLI takes precedence means the other origin messages will have
+  no effect.
+
 ### Transactionality of Sets with multiple Origins
 
 Where a `SetRequest` specifies more than one `origin` - i.e., two or more
@@ -112,7 +130,7 @@ an error status returned upon responding to the `Set` RPC.
 
 ### Example: OpenConfig and CLI Data
 
-If a client wishes to replace both OpenConfig and CLI-modelled data
+If a client wishes to replace OpenConfig and update CLI-modelled data
 concurrently, it can send the following `SetRequest`:
 
 ```
@@ -138,7 +156,7 @@ replace: <
     `
   >
 >
-replace: <
+update: <
   path: <
     origin: "cli"
   >
@@ -148,10 +166,14 @@ replace: <
 >
 ```
 
-This transaction replaces the contents of both the "openconfig" and "cli"
-origins.  The first `replace` message replaces the OpenConfig origin at the root
-(as specified by the zero-length array of `PathElem` messages) to the specified
-JSON. The second `replace` replaces the CLI configuration of the device is
-replaced with the string specified in the `ascii_val` field. Per the gNMI
-specification, both `replace` operations MUST successfully be applied, otherwise
-the configuration change should be rolled back.
+This transaction replaces the contents of the "openconfig" origins and then
+updates the configuration with CLI content.  The first `replace` message
+replaces the OpenConfig origin at the root (as specified by the zero-length
+array of `PathElem` messages) to the specified JSON. The `update` operation
+adds the CLI configuration of the device with with the string specified in the
+`ascii_val` field. Per the gNMI specification, both operations MUST
+successfully be applied, otherwise the configuration change should be rolled
+back.
+
+The expected resulting configuration will contain an enabled interface and a
+router bgp 15169 statement.
--- a/rpc/gnmi/mixed-schema.md
+++ b/rpc/gnmi/mixed-schema.md
@@ -114,10 +114,16 @@ Special considerations for mixing native schema are currently undefined.
 * The `replace` operation for a setRequest containing only `origin` `cli` MUST
   be treated as a full device configuration replacement.
 * The `replace` operation for a setRequest containing both CLI and OpenConfig,
-  the OpenConfig and CLI should be merged, with precedence to CLI for any
-  overlapping configuration, and then the replace operation performed.  The
-  goal is to allow configuration replacement to occur at a sub-tree level using
-  both OpenConfig and the CLI configuration data.
+  the OpenConfig and CLI SHOULD be supported.  The content of the two origins
+  SHOULD be be merged, with precedence to CLI for any overlapping
+  configuration, and then the replace operation performed.  Since the CLI is
+  interpreted as a full configuration replacement, the result of this
+  setRequest should be a full configuration replacement with the merged
+  content of the CLI and OpenConfig paths.
+* A `replace` operation for `origin` `openconfig`, followed by a `update`
+  operation for `origin` `cli` in a single setRequest SHOULD be supported.
+  The goal is to allow configuration replacement to occur at a sub-tree level
+  using both OpenConfig and the CLI configuration data.
 
 ### Transactionality of Sets with multiple Origins
 
--- a/rpc/gnmi/mixed-schema.md
+++ b/rpc/gnmi/mixed-schema.md
@@ -1,10 +1,7 @@
 # Mixing Schemas in gNMI
-**Contributors**: aashaikh, hines, robjs, csl  
-October 2016, Updated May 2018  
-*Implementation Status*: Merged
-
 
 ## Problem Statement
+
 Today's network devices support multiple schema representations of their
 underlying configuration and telemetry databases. Particularly, devices
 typically support:
@@ -44,14 +41,14 @@ within the message is uniquely identified by the tuple of `<origin, path>`.
 The `origin` field is valid in any context of a `Path` message. Typically it is
 used:
 
- * In a `SetRequest` to indicate a particular schema is being used to modify
-   the target configuration.
- * In a `GetRequest` to retrieve the contents of a particular schema, or in
-   a `GetResponse` to indicate that the payload contains data from a
-   particular `<origin, path>` schema.
- * In a `SubscribeRequest` to subscribe to paths within a particular schema,
-   or `SubscribeResponse` to indicate an update corresponds to a particular
-   `<origin, path>` tuple.
+* In a `SetRequest` to indicate a particular schema is being used to modify
+  the target configuration.
+* In a `GetRequest` to retrieve the contents of a particular schema, or in
+  a `GetResponse` to indicate that the payload contains data from a
+  particular `<origin, path>` schema.
+* In a `SubscribeRequest` to subscribe to paths within a particular schema,
+  or `SubscribeResponse` to indicate an update corresponds to a particular
+  `<origin, path>` tuple.
 
 If more than one `origin` is to be used within any message, a path in the
 `prefix` MUST NOT be specified, since a prefix applies to all paths within the
@@ -100,22 +97,27 @@ constrained to the specified origin. Particularly:
   MUST specify multiple paths within the `delete` of the `SetRequest`.
 
 ### Special considerations for the use of Origin CLI
-* If a `Set` RPC contains `origin` CLI, the CLI origin MUST be given precedence
-  over the OpenConfig origin.  This precedence is independent of the order in
-  which origin CLI appears.  This overrides the gNMI specification requirement
-  of ordered path based precedence within a `SetRequest`.  The goal of this
-  precedence is to create deterministic behavior for resolving overlapping
-  configuration.
-* The `update` operation MUST be supported for `origin` CLI.  The `update`
-  operation is compatible with the requirement for `origin` CLI to take
-  precedence.
-  
-* Because there is no explicit expectation of CLI to use a path based
-  structure, use of the `replace` operation is not recommended unless the
-  intent is to replace the entire configuration with the contents of the origin CLI message.  Replacing
-  the entire configuration with CLI content in a mixed origin `SetRequest`
-  where origin CLI takes precedence means the other origin messages will have
-  no effect.
+
+The following rules apply to a setRequest containing `origin` CLI.  TODO:
+Special considerations for mixing native schema are currently undefined.
+
+* If a `SetRequest` contains `origin` `cli` and `origin` `openconfig`, the CLI
+  MUST be given precedence over the OpenConfig origin.  This precedence is
+  independent of the order in which origin CLI appears.  This overrides the
+  gNMI specification requirement of ordered path based precedence within a
+  `SetRequest`.  The goal of this precedence is to create deterministic
+  behavior for resolving overlapping configuration.  For setRequests without
+  `origin` CLI, order based precedence still applies.
+* The `update` operation MUST be supported for setRequests containing `origin`
+  `cli` and `origin` `openconfig`.  The `update` operation is compatible with
+  the requirements for `origin` `cli` to take precedence.
+* The `replace` operation for a setRequest containing only `origin` `cli` MUST
+  be treated as a full device configuration replacement.
+* The `replace` operation for a setRequest containing both CLI and OpenConfig,
+  the OpenConfig and CLI should be merged, with precedence to CLI for any
+  overlapping configuration, and then the replace operation performed.  The
+  goal is to allow configuration replacement to occur at a sub-tree level using
+  both OpenConfig and the CLI configuration data.
 
 ### Transactionality of Sets with multiple Origins
 


rpc/gnmi/mixed-schema.md
```diff
@@ -99,6 +99,24 @@ constrained to the specified origin. Particularly:
   the specified `origin`. To delete contents from multiple origins, a client
   MUST specify multiple paths within the `delete` of the `SetRequest`.
 
+### Special considerations for the use of Origin CLI
+* If a `Set` RPC contains `origin` CLI, the CLI origin MUST be given precedence
+  over the OpenConfig origin.  This precedence is independent of the order in
```
hellt, line 104: @dplore shall we further clarify the order precedence for native/default origins?
Say you have a switch that supports the following origins:
1. `cli`
2. no-origin - meaning default origin that may be openconfig or vendor-native schema
3. `oc`

Do we want to say here that cli origin takes precedence over any other potentially available origins (like sonic may have some orthodox `config-db` origin, for instance) and then the rest of the origins are processed in the order-based fashion?

rpc/gnmi/mixed-schema.md
```diff
@@ -138,7 +156,7 @@ replace: <
     `
   >
 >
-replace: <
+update: <
```
rpazyaquian, line 159: What happens if I make a set request with two `update`s of origin `cli`?

rpc/gnmi/mixed-schema.md
```diff
@@ -99,6 +99,24 @@ constrained to the specified origin. Particularly:
   the specified `origin`. To delete contents from multiple origins, a client
   MUST specify multiple paths within the `delete` of the `SetRequest`.
 
+### Special considerations for the use of Origin CLI
+* If a `Set` RPC contains `origin` CLI, the CLI origin MUST be given precedence
```
dan-lepage, line 103: Just to be clear, "precedence" in this case means not that the CLI commands should precede the OC updates, but rather that any value changed by both a CLI command and an OC update should take on the value set by the CLI command?

And is this the same as saying "all OC updates should be applied first, then all CLI updates",  i.e. the sequence

    cli command 1
    oc update 1
    cli command 2
    oc update 2

should behave exactly equivalently to

    oc update 1
    oc update 2
    cli command 1
    cli command 2
?

rpc/gnmi/mixed-schema.md
robshakir: @hellt -- I think (as specified in this doc) "" implies origin `openconfig` I believe. Is your implication that `openconfig` (not `oc`) is different to `""` above?

robshakir, line 104: Over any other origin? Remember that there are origins that can be defined by the vendor, or other projects. I believe there are origins used for elements of P4 information (or there were intended to be previously).

rpc/gnmi/mixed-schema.md
```diff
@@ -99,6 +99,24 @@ constrained to the specified origin. Particularly:
...
+  over the OpenConfig origin.  This precedence is independent of the order in
+  which origin CLI appears.  This overrides the gNMI specification requirement
+  of ordered path based precedence within a `SetRequest`.  The goal of this
+  precedence is to create deterministic behavior for resolving overlapping
+  configuration.
```
robshakir, line 108: Clarify that for origins _other_ than `cli` this rule continues to be implemented.

rpc/gnmi/mixed-schema.md
```diff
@@ -99,6 +99,24 @@ constrained to the specified origin. Particularly:
...
+* The `update` operation MUST be supported for `origin` CLI.  The `update`
+  operation is compatible with the requirement for `origin` CLI to take
+  precedence.
+  
+* Because there is no explicit expectation of CLI to use a path based
+  structure, use of the `replace` operation is not recommended unless the
+  intent is to replace the entire configuration with the contents of the origin CLI message.  Replacing
+  the entire configuration with CLI content in a mixed origin `SetRequest`
+  where origin CLI takes precedence means the other origin messages will have
+  no effect.
```
robshakir, line 118: Is this correct?

If I send a replace for origin cli, and then openconfig afterwards, does this mean that there is really only CLI even where it does not overlap?

If this is the case I do not think this is intended behaviour - because then there is no way to ever remove something that is already in the CLI without explicitly deleting.

I think the precedence applies only when there is overlap of elements.

rpc/gnmi/mixed-schema.md
robshakir, line 108: Clarify that for origins _other_ than `cli` this rule continues to be implemented.

rpc/gnmi/mixed-schema.md
robshakir: What I mean here.

If you have the initial state:

```
origin: cli contents: "foobar"
```

and then apply

```
update origin: cli contents: "baz"
update origin: openconfig contents `{"bat": "foo"}`
```

the cli contents "foobar` are never removed, and so at this point one cannot ever do a full configuration replace.

I think you need to rather allow `replace` of the CLI, and say that as long as "bat` does not overlap with "baz" in the CLI, then both are merged.

rpc/gnmi/mixed-schema.md
```diff
@@ -148,10 +166,14 @@ replace: <
...
 
-This transaction replaces the contents of both the "openconfig" and "cli"
-origins.  The first `replace` message replaces the OpenConfig origin at the root
-(as specified by the zero-length array of `PathElem` messages) to the specified
-JSON. The second `replace` replaces the CLI configuration of the device is
-replaced with the string specified in the `ascii_val` field. Per the gNMI
-specification, both `replace` operations MUST successfully be applied, otherwise
-the configuration change should be rolled back.
+This transaction replaces the contents of the "openconfig" origins and then
+updates the configuration with CLI content.  The first `replace` message
```
robshakir, line 170: This still references `replace` on the CLI -- which I think is correct, but you disallowed it above :-)

robshakir (changes requested):

rpc/gnmi/mixed-schema.md
sebsapa: `replace origin=cli` should be treated as a full device config replace.
As such, if `origin=openconfig` is provided it should be specified with a path at root level; otherwise reject it.

I can see this being useful for operators transitioning to use OpenConfig and start moving `cli` towards `openconfig`.

`cli` could also take precedence here if there is a conflict.

rpc/gnmi/mixed-schema.md
dplore: I would like to defer the decision of how to deal with 3 origins (OC, Vendor native yang and CLI)  and keep this PR to only OpenConfig (which is also known as "") and CLI origin.    This essentially means at the moment, that there is not a gNMI specification for how to mix vendor native, openconfig and CLI in the same setRequest.  There are other ways for an operator to get deterministic behavior though, by performing separate setrequests for example.

rpc/gnmi/mixed-schema.md
dplore: Precedence is intended to be interpreted as you described.  

Yes, a setRequest with either of the above sequences of updates should have the same effect on the device configuration.

rpc/gnmi/mixed-schema.md
dplore: I would like to contain the scope of this clarification to setRequest containing a mix of OpenConfig and CLI origins.   

In a future PR, we could consider additional combinations.  Please feel free to reopen this thread if you think this will cause a conflict or leads to some invalidation of the OpenConfig + CLI behavior defined here.

rpc/gnmi/mixed-schema.md
dplore: added

rpc/gnmi/mixed-schema.md
dplore: I have added an update to clarify the replace behavior for mixed OC + CLI.  I appreciate your feedback.

rpc/gnmi/mixed-schema.md
dplore: I think I resolved this conflict now.  :)

rpc/gnmi/mixed-schema.md
```diff
@@ -99,6 +96,35 @@ constrained to the specified origin. Particularly:
...
+  the OpenConfig and CLI SHOULD be supported.  The content of the two origins
+  SHOULD be be merged, with precedence to CLI for any overlapping
+  configuration, and then the replace operation performed.  Since the CLI is
+  interpreted as a full configuration replacement, the result of this
+  setRequest should be a full configuration replacement with the merged
+  content of the CLI and OpenConfig paths.
+* A `replace` operation for `origin` `openconfig`, followed by a `update`
+  operation for `origin` `cli` in a single setRequest SHOULD be supported.
+  The goal is to allow configuration replacement to occur at a sub-tree level
+  using both OpenConfig and the CLI configuration data.
```
dplore, line 126: I would like to highlight that this scenario is the highest priority use case for Google.  IMO, this scenario provides the most flexibility to replace the portions of the configuration needed (include the full configuration if desired), while mixing in CLI to cover gaps.   We have operational use cases where replacement of subtrees is necessary.

rpc/gnmi/mixed-schema.md
dplore: This is currently undefined.  To obtain defined behavior, I suggest combining the single setRequest with 2 or more CLI updates into a single setRequest with a single CLI update.

rpc/gnmi/mixed-schema.md
```diff
@@ -99,6 +96,35 @@ constrained to the specified origin. Particularly:
...
   MUST specify multiple paths within the `delete` of the `SetRequest`.
 
+### Special considerations for the use of Origin CLI
+
+The following rules apply to a setRequest containing `origin` CLI.  TODO:
+Special considerations for mixing native schema are currently undefined.
+
+* If a `SetRequest` contains `origin` `cli` and `origin` `openconfig`, the CLI
+  MUST be given precedence over the OpenConfig origin.  This precedence is
+  independent of the order in which origin CLI appears.  This overrides the
```
shichuzhu, line 106: Instead of "This precedence is independent of the order in which origin CLI appears",
how about saying "It's invalid for a SetRequest to contain oc/cli origins that do not fall into the following categories...", which I think has the benefit
1. forces the user to explicitly think of their expectation.
2. makes it easier to debug (an grpc error message is better than verifying device behavior after an 'accepted' SetRequest).

rpc/gnmi/mixed-schema.md
```diff
@@ -99,6 +96,35 @@ constrained to the specified origin. Particularly:
   the specified `origin`. To delete contents from multiple origins, a client
   MUST specify multiple paths within the `delete` of the `SetRequest`.
 
+### Special considerations for the use of Origin CLI
+
+The following rules apply to a setRequest containing `origin` CLI.  TODO:
```
hellt, line 101:
```suggestion
The following rules apply to a `SetRequest` containing `origin` CLI.  TODO:
```

rpc/gnmi/mixed-schema.md
```diff
@@ -99,6 +96,35 @@ constrained to the specified origin. Particularly:
...
+  the requirements for `origin` `cli` to take precedence.
+* The `replace` operation for a setRequest containing only `origin` `cli` MUST
+  be treated as a full device configuration replacement.
+* The `replace` operation for a setRequest containing both CLI and OpenConfig,
+  the OpenConfig and CLI SHOULD be supported.  The content of the two origins
+  SHOULD be be merged, with precedence to CLI for any overlapping
+  configuration, and then the replace operation performed.  Since the CLI is
+  interpreted as a full configuration replacement, the result of this
+  setRequest should be a full configuration replacement with the merged
+  content of the CLI and OpenConfig paths.
```
hellt, line 122: @dplore can you help me understand this use case better? For a SetRequest with `replace` op that contains both `cli` and `openconfig` origins, would the following order of operations match what is described above?

1. a candidate datastore is created
2. first, replaces of openconfig origin apply
3. then CLI commands apply on top of the changes that were made in 2 to make sure that CLI has precedence for the overlapping changes
4. replace happens at root (courtesy of CLI origin behavior) and whatever was a result of 2+3 is applied

rpc/gnmi/mixed-schema.md
```diff
@@ -99,6 +96,35 @@ constrained to the specified origin. Particularly:
   the specified `origin`. To delete contents from multiple origins, a client
   MUST specify multiple paths within the `delete` of the `SetRequest`.
 
+### Special considerations for the use of Origin CLI
```
hellt, line 99: Since `cli` origin is quite special, should we provide guidance on how it is meant to be used outside of SetRequest which is described below?

What happens when `cli` origin is used in a get request? I'd assume that for NOSes that maintain different datastores a GetRequest should only contain CLI commands that can be executed from the  equivalent of a `state` datastore.
But what if someone uses a GetRequest with configuration CLI commands inside of it?

Any guidance on what should be returned back if CLI origin is used in a Subscribe request?

hellt: Also I see the following sentence

```
Where the origin field is specified as cli, the path should be ignored and the value specified within the Update considered as CLI input.
```

which implies `cli` origin can only be used in a Set Requests. If that is true, then maybe we should put an accent on that.

rpc/gnmi/mixed-schema.md
dplore: gNMI specifically requires order based precedence.  This special case for origin CLI overrides this by asserting CLI always takes precedence over OC, regardless of order.

rpc/gnmi/mixed-schema.md
dplore: Yes, this is one way to implement the desired behavior for a setRequest containing: 
```
replace origin openconfig
update origin CLI
```

rpc/gnmi/mixed-schema.md
dplore: I am open to feedback on the topic of getRequest for origin CLI.  My first thought is a getRequest for origin CLI should return the complete config in the vendor's chosen 'CLI' format.

rpc/gnmi/mixed-schema.md
```diff
@@ -99,6 +96,35 @@ constrained to the specified origin. Particularly:
...
+The following rules apply to a setRequest containing `origin` CLI.  TODO:
+Special considerations for mixing native schema are currently undefined.
+
+* If a `SetRequest` contains `origin` `cli` and `origin` `openconfig`, the CLI
+  MUST be given precedence over the OpenConfig origin.  This precedence is
+  independent of the order in which origin CLI appears.  This overrides the
+  gNMI specification requirement of ordered path based precedence within a
+  `SetRequest`.  The goal of this precedence is to create deterministic
+  behavior for resolving overlapping configuration.  For setRequests without
+  `origin` CLI, order based precedence still applies.
```
jsterne, line 110: This last sentence seems to somewhat contradict what we converged on for #474/#740. I thought that it wasn't so much about origin = CLI as origin = *any native schema* (CLI or modelled)?

earies: +1 to @jsterne comment above - what `cli` really means is "native schema" + ascii_val encoding.  I've always been of the opinion it should be named as such to allow for multiple encodings of a native schema.  `cli` is an improper origin name especially in comparison to `openconfig` which can encode data per the appropriate TypedVal

But the wording above implies that if origin=openconfig + native modeled data as say JSON_IETF (which would fall into another origin) loops right back around to the same problem

rpc/gnmi/mixed-schema.md
sebsapa: Arista supports getRequest with cli origin as a way to retrieve state. Operators can use this a way to inspect cli configuration, or any other cli state.

For example:
```
gnmi  get origin=cli "show version"
/show version:
{
  "version": "1.2345"
}
```

rpc/gnmi/mixed-schema.md
```diff
@@ -99,6 +96,35 @@ constrained to the specified origin. Particularly:
...
+* The `replace` operation for a setRequest containing only `origin` `cli` MUST
+  be treated as a full device configuration replacement.
+* The `replace` operation for a setRequest containing both CLI and OpenConfig,
+  the OpenConfig and CLI SHOULD be supported.  The content of the two origins
+  SHOULD be be merged, with precedence to CLI for any overlapping
+  configuration, and then the replace operation performed.  Since the CLI is
+  interpreted as a full configuration replacement, the result of this
+  setRequest should be a full configuration replacement with the merged
+  content of the CLI and OpenConfig paths.
+* A `replace` operation for `origin` `openconfig`, followed by a `update`
```
nkitchen, line 123:
```suggestion
* A `replace` operation for `origin` `openconfig`, followed by an `update`
```

rpc/gnmi/mixed-schema.md
```diff
@@ -99,6 +96,35 @@ constrained to the specified origin. Particularly:
...
+  behavior for resolving overlapping configuration.  For setRequests without
+  `origin` CLI, order based precedence still applies.
+* The `update` operation MUST be supported for setRequests containing `origin`
+  `cli` and `origin` `openconfig`.  The `update` operation is compatible with
+  the requirements for `origin` `cli` to take precedence.
+* The `replace` operation for a setRequest containing only `origin` `cli` MUST
+  be treated as a full device configuration replacement.
+* The `replace` operation for a setRequest containing both CLI and OpenConfig,
+  the OpenConfig and CLI SHOULD be supported.  The content of the two origins
+  SHOULD be be merged, with precedence to CLI for any overlapping
```
nkitchen, line 118:
```suggestion
  SHOULD be merged, with precedence to CLI for any overlapping
```

rpc/gnmi/mixed-schema.md
```diff
@@ -99,6 +96,35 @@ constrained to the specified origin. Particularly:
...
+  `SetRequest`.  The goal of this precedence is to create deterministic
+  behavior for resolving overlapping configuration.  For setRequests without
+  `origin` CLI, order based precedence still applies.
+* The `update` operation MUST be supported for setRequests containing `origin`
+  `cli` and `origin` `openconfig`.  The `update` operation is compatible with
+  the requirements for `origin` `cli` to take precedence.
+* The `replace` operation for a setRequest containing only `origin` `cli` MUST
+  be treated as a full device configuration replacement.
+* The `replace` operation for a setRequest containing both CLI and OpenConfig,
+  the OpenConfig and CLI SHOULD be supported.  The content of the two origins
```
nkitchen, line 117:
```suggestion
* The `replace` operation for a setRequest containing both CLI and OpenConfig
  SHOULD be supported.  The content of the two origins
```

rpc/gnmi/mixed-schema.md
hellt: > I am open to feedback on the topic of getRequest for origin CLI. My first thought is a getRequest for origin CLI should return the complete config in the vendor's chosen 'CLI' format.

oh, I like that. e.g. get with origin `cli` and empty path returns a complete config in the CLI format, and if a path is specified, if the NOS supports it, it returns the CLI config of a path subtree

rpc/gnmi/mixed-schema.md
```diff
@@ -99,6 +96,35 @@ constrained to the specified origin. Particularly:
...
+  gNMI specification requirement of ordered path based precedence within a
+  `SetRequest`.  The goal of this precedence is to create deterministic
+  behavior for resolving overlapping configuration.  For setRequests without
+  `origin` CLI, order based precedence still applies.
+* The `update` operation MUST be supported for setRequests containing `origin`
+  `cli` and `origin` `openconfig`.  The `update` operation is compatible with
+  the requirements for `origin` `cli` to take precedence.
+* The `replace` operation for a setRequest containing only `origin` `cli` MUST
+  be treated as a full device configuration replacement.
+* The `replace` operation for a setRequest containing both CLI and OpenConfig,
```
brwallis, line 116: The wording here (and throughout) implies that implementations model each origin as it's own DB - is that really the expectation? For example if I create object `A` that exists only in `cli` and then do a replace on root in `openconfig` should object `A` still exist? https://github.com/openconfig/reference/blob/5038bb2d94444da0fe3a679ce902b8ca491b2e95/rpc/gnmi/mixed-schema.md?plain=1#L91 implies that the two shouldn't interact but it seems indirectly that they must if they are both modeling an underlying object that both can manipulate. It's not clear what the second SetRequest from the below example would look like, but would it need to continue to send the `update` portion for the `cli` origin?

To add another example, if I create object `B` in `cli` with field `F` and object `B` IS modeled in `openconfig`, without visibility to field `F`, what happens if I were to do a replace on object `B` in `openconfig`?

Now what if the path being replaced were root in both cases (implied with the `cli` origin)? Doesn't the implied replace on root in `cli` essentially make it's full contents win in a merge between the two origins?

dplore: > The wording here (and throughout) implies that implementations model each origin as it's own DB - is that really the expectation? For example if I create object `A` that exists only in `cli` and then do a replace on root in `openconfig` should object `A` still exist?

If object A is not contained in the openconfig replace at the root, the object A should be removed.  If object A is contained in the openconfig replace at the root, object A should be replaced with the new value.
 
> https://github.com/openconfig/reference/blob/5038bb2d94444da0fe3a679ce902b8ca491b2e95/rpc/gnmi/mixed-schema.md?plain=1#L91
> 
> implies that the two shouldn't interact but it seems indirectly that they must if they are both modeling an underlying object that both can manipulate. It's not clear what the second SetRequest from the below example would look like, but would it need to continue to send the `update` portion for the `cli` origin?

Thanks for catching this.  I believe the text you've linked to here needs to be updated or removed as part of this pull request as it is in conflict with the use of mixed OpenConfig and CLI.    A setrequest containing origin OC must update/replace configuration which is also modeled by CLI.  Presumably 100% of all configuration items are modeled by CLI and OpenConfig is in reality, the subset.

> To add another example, if I create object `B` in `cli` with field `F` and object `B` IS modeled in `openconfig`, without visibility to field `F`, what happens if I were to do a replace on object `B` in `openconfig`?
> 

If the setrequest only contains origin `openconfig` and replaces object `B`, then object `B` should be replaced.

If the setrequest replace contains origin `openconfig` and origin `cli` and both reference object `B`, the `cli` reference should override or take precedence over the `openconfig` reference.  

> Now what if the path being replaced were root in both cases (implied with the `cli` origin)? Doesn't the implied replace on root in `cli` essentially make it's full contents win in a merge between the two origins?

Yes, I think this is why a setrequest with `replace` origin `openconfig` and `replace` origin `cli` in a single setrequest is not a useful operation.  The `cli` replace will take precedence rendering the `openconfig` origin data useless.   

Instead I believe the useful pattern is a setrequest with `replace` origin `openconfig` and `update` origin `cli` is in the example now.  If the `replace` origin `openconfig` contains a root path, then the full configuration is replaced with the contents of the openconfig, followed by an `update` of the origin `cli` as per [gNMI 3.4](https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-specification.md#34-modifying-state)

rpc/gnmi/mixed-schema.md
dplore: +1 to @hellt 

The topic of this document however doesn't quite match with this information.  If @sebsapa interpretation is not intuitive and/or we feel the need to make it explicit, perhaps we should start a separate issue/PR.

rpc/gnmi/mixed-schema.md
dplore: I think we want to explicitly maintain support for origin `cli`.    We could elaborate on the definition of what origin `cli` means in a separate pull request to update  [gNMI 2.2.2](https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-specification.md#222-paths) where the data types and origins are defined.

dplore: To answer @jsterne question, the intention is this section only applies to origin `cli`.   The definition of origin `cli` is admittedly not very strictly defined.

rpc/gnmi/mixed-schema.md
```diff
@@ -99,6 +96,35 @@ constrained to the specified origin. Particularly:
...
+  MUST be given precedence over the OpenConfig origin.  This precedence is
+  independent of the order in which origin CLI appears.  This overrides the
+  gNMI specification requirement of ordered path based precedence within a
+  `SetRequest`.  The goal of this precedence is to create deterministic
+  behavior for resolving overlapping configuration.  For setRequests without
+  `origin` CLI, order based precedence still applies.
+* The `update` operation MUST be supported for setRequests containing `origin`
+  `cli` and `origin` `openconfig`.  The `update` operation is compatible with
+  the requirements for `origin` `cli` to take precedence.
+* The `replace` operation for a setRequest containing only `origin` `cli` MUST
```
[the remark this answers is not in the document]
dplore: Separate set requests should behave as defined by gNMI.   A setrequest with origin `openconfig` should update or replace the referenced "configuration item(s)" regardless of what setrequest and origin was used in the past.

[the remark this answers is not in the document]
dplore: Precedence in this case is determined by use of an `update` or `replace` operation and of course, the ordering of the setrequests.  Each setrequest should be processed as a transaction, separately and atomically.

rpc/gnmi/mixed-schema.md
jsterne: It feels like this "OpenConfig replace" + "Native update" should be defined to work (in a similar way) whether Native CLI or Native YANG modelled is used. At least for the basic concepts like having the native take precedence,

rpc/gnmi/mixed-schema.md
```diff
@@ -44,14 +41,14 @@ within the message is uniquely identified by the tuple of `<origin, path>`.
...
- * In a `SetRequest` to indicate a particular schema is being used to modify
-   the target configuration.
- * In a `GetRequest` to retrieve the contents of a particular schema, or in
-   a `GetResponse` to indicate that the payload contains data from a
-   particular `<origin, path>` schema.
- * In a `SubscribeRequest` to subscribe to paths within a particular schema,
-   or `SubscribeResponse` to indicate an update corresponds to a particular
-   `<origin, path>` tuple.
+* In a `SetRequest` to indicate a particular schema is being used to modify
+  the target configuration.
```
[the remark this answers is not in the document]
dplore: @earies please suggest the relevant schemas and encodings for JunOS (set request, curly brace and junos yang) be indicated in the origin and encoding.

rpc/gnmi/mixed-schema.md
```diff
@@ -44,14 +41,14 @@ within the message is uniquely identified by the tuple of `<origin, path>`.
...
+* In a `GetRequest` to retrieve the contents of a particular schema, or in
+  a `GetResponse` to indicate that the payload contains data from a
+  particular `<origin, path>` schema.
+* In a `SubscribeRequest` to subscribe to paths within a particular schema,
+  or `SubscribeResponse` to indicate an update corresponds to a particular
+  `<origin, path>` tuple.
```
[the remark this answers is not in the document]
dplore: I agree, I will add this clarification for the `cli` case.

rpc/gnmi/mixed-schema.md
```diff
@@ -99,6 +96,35 @@ constrained to the specified origin. Particularly:
...
+  independent of the order in which origin CLI appears.  This overrides the
+  gNMI specification requirement of ordered path based precedence within a
+  `SetRequest`.  The goal of this precedence is to create deterministic
+  behavior for resolving overlapping configuration.  For setRequests without
+  `origin` CLI, order based precedence still applies.
+* The `update` operation MUST be supported for setRequests containing `origin`
+  `cli` and `origin` `openconfig`.  The `update` operation is compatible with
+  the requirements for `origin` `cli` to take precedence.
+* The `replace` operation for a setRequest containing only `origin` `cli` MUST
+  be treated as a full device configuration replacement.
```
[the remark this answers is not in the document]
dplore: Yes, this assumes that to be useful, the `replace` operator for `origin` `cli` should cover all functionality of the device.  Is this a rhetorical comment or does JunOS not support all functions through it's "set" and {} curly brace syntax?

rpc/gnmi/mixed-schema.md
jsterne: To be more concrete: I feel pretty strongly that this sentence should not be part of the spec: "For setRequests without `origin` CLI, order based precedence still applies."

jsterne:
One consequence of origin 'CLI' being a mix of:
1) schema, and
2) instance data format/encoding
is potential confusion when OpenConfig data is sent using CLI format. (is that OpenConfig or is it CLI?).

IMO the "precedence" rule of native over OpenConfig should be about native *schema* taking precedence over OpenConfig schema (regardless of format/encoding).

Consider a single Set request with the following (in this order):
1) update with origin 'CLI', but the data contains purely OpenConfig data (in CLI format)
2) update with origin 'OpenConfig'
[and there is overlap/conflict between (1) and (2), e.g. (1) sets mtu to 1504 and (2) sets mtu to 1492]

By the current text it would mandate that the configuration in (1) take precedence. But that seems to somewhat conflict with the general gNMI 'Set' approach of processing the updates in order.

rpc/gnmi/mixed-schema.md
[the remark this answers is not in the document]
earies: JUNOS does support this but I raise this abstract of the implementation I'm associated with because it was never the intention of OpenConfig modeling to actually be exposed in vendor's CLIs - those that have done so, likely have due to already existing architectural principles and byproducts but this is not the case for all implementations.

With this verbiage, I'm afraid that 3rd party schema coverage for vendor CLIs is then implicitly mandated all of which will have different user experiences/representations.  Those that do not support now need to in order to comply or we have diverged implementations and thus behaviors of a specification that cannot be conformed to.  Essentially this silently raises the barrier of entry.

I personally think mandating CLI behaviors is scope creep at this stage.  In addition there are already other behaviors occurring that are starting to exclude other NBI that have me seriously concerned.

rpc/gnmi/mixed-schema.md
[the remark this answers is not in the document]
earies: Thx @dplore - I'll leave this to another PR/issue to track but want to mention that a somewhat parallel issue is also seen that while the `Capabilities` RPC reports "supported encodings" I think we will want to clarify that this type of reporting is mostly meaningless as each RPC (at least in the context of gNMI) cannot support nor _should_ support all encodings for all operations - its similar in the fact that we have per RPC/origin/encoding behaviors coming out of this that need to be carefully documented.

rpc/gnmi/mixed-schema.md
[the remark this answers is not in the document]
robshakir: @earies - I'm not sure that I follow your comment here. Origin `cli` was defined to mean "some out of band agreed ASCII format that the device supports". In this case, for JUNOS it could be either the `set xxx` or your `{}` format -- but isn't that something that you're agreeing outside of the concept of this "special" origin?

Maybe the concern here is the word `schema` in the sentence above? Should we rather use `tree`? I think of the different origins as different trees that define some configuration / telemetry representation on the device.

rpc/gnmi/mixed-schema.md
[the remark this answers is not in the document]
robshakir: One thing to think about: this means that you must support a `GetRequest` for the origin CLI, which means no chunking of the configuration into separate gRPC messages. Is this desirable?

It seems like we might want to allow this to be something that COULD be implemented in the spec, but not require it for an initial implementation (i.e., it's a MAY).

rpc/gnmi/mixed-schema.md
```diff
@@ -99,6 +96,35 @@ constrained to the specified origin. Particularly:
   the specified `origin`. To delete contents from multiple origins, a client
   MUST specify multiple paths within the `delete` of the `SetRequest`.
 
+### Special considerations for the use of Origin CLI
+
+The following rules apply to a `SetRequest` containing `origin` CLI.  TODO:
+Special considerations for mixing native schema are currently undefined.
```
robshakir, line 102: Please don't leave a TODO like this. It's pretty unclear what this sentence means.

What does "mixing native schema" mean? If there's something specific that we are not discussing then that's fine (mixing a native YANG-modelled schema with a non-native one? overlapping contents between two YANG-modelled schemas?), but this sentence doesn't make a whole lot of sense.

rpc/gnmi/mixed-schema.md
robshakir, line 99: Do you want to make this "special considerations for the use of origin `openconfig` and origin `cli`"? This seems to be the really specific case that we're handling.

rpc/gnmi/mixed-schema.md
```diff
@@ -99,6 +96,35 @@ constrained to the specified origin. Particularly:
...
+
+* If a `SetRequest` contains `origin` `cli` and `origin` `openconfig`, the CLI
+  MUST be given precedence over the OpenConfig origin.  This precedence is
+  independent of the order in which origin CLI appears.  This overrides the
+  gNMI specification requirement of ordered path based precedence within a
+  `SetRequest`.  The goal of this precedence is to create deterministic
+  behavior for resolving overlapping configuration.  For setRequests without
```
robshakir, line 109:
```suggestion
  behavior for resolving overlapping configuration.  For `SetRequest` messages without
```

rpc/gnmi/mixed-schema.md
```diff
@@ -99,6 +96,35 @@ constrained to the specified origin. Particularly:
...
+  `origin` CLI, order based precedence still applies.
+* The `update` operation MUST be supported for setRequests containing `origin`
```
robshakir, line 111:
```suggestion
* The `update` operation MUST be supported for SetRequests containing `origin`
```

rpc/gnmi/mixed-schema.md
```diff
@@ -99,6 +96,35 @@ constrained to the specified origin. Particularly:
...
+  `cli` and `origin` `openconfig`.  The `update` operation is compatible with
+  the requirements for `origin` `cli` to take precedence.
+* The `replace` operation for a setRequest containing only `origin` `cli` MUST
```
robshakir, line 114:
```suggestion
* The `replace` operation for a SetRequest containing only `origin` `cli` MUST
```

rpc/gnmi/mixed-schema.md
```diff
@@ -99,6 +96,35 @@ constrained to the specified origin. Particularly:
...
+  be treated as a full device configuration replacement.
+* The `replace` operation for a setRequest containing both CLI and OpenConfig
```
robshakir, line 116:
```suggestion
* The `replace` operation for a SetRequest containing both CLI and OpenConfig
```

rpc/gnmi/mixed-schema.md
```diff
@@ -99,6 +96,35 @@ constrained to the specified origin. Particularly:
...
+  SHOULD be supported.  The content of the two origins
+  SHOULD be merged, with precedence to CLI for any overlapping
+  configuration, and then the replace operation performed.  Since the CLI is
+  interpreted as a full configuration replacement, the result of this
+  setRequest should be a full configuration replacement with the merged
+  content of the CLI and OpenConfig paths.
+* A `replace` operation for `origin` `openconfig`, followed by an `update`
+  operation for `origin` `cli` in a single setRequest SHOULD be supported.
```
robshakir, line 124:
```suggestion
  operation for `origin` `cli` in a single SetRequest SHOULD be supported.
```

robshakir (changes requested):

robshakir:
Is there a simplification that we can write here that meets the same requirement whilst not having to deal with many edge cases that seem irrelevant?

1. We support only `replace` for origin `cli`.
2. `replace` with origin `cli` MUST be first in the `SetRequest`.
3. Application of another schema -- OpenConfig/vendor native models is then applied on top of this new candidate that was formed by the replace of the CLI. (i.e., replace `cli` means "replace everything on the device even if it's not in this tree").

This gives us a much simpler solution - that makes not just the `origin: "cli"` specific but also the semantics of a full replace.

rpc/gnmi/mixed-schema.md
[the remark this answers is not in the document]
dplore: When a gNMI setrequest origin OC sets a new ip address on an interface, the new ip address should take effect regardless of what method was previously used.

rpc/gnmi/mixed-schema.md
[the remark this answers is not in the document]
robshakir: This seems super-complicated to me -- especially concepts like `cli_set_val`.

I'd suggest we come back to what we feel is reasonable to support given existing use cases that are being proposed. Otherwise, we're going to end up trying to figure out how to support many different combinations that there will be no real-use-case test coverage for because they are not operationally interesting.

I've documented a simpler proposal that I think gives us some room to simplify below.

rpc/gnmi/mixed-schema.md
dplore: I can remove that sentence and leave that topic open/unclarified for now.  We should not bring in vendor yang mixing issues in this PR.

rpc/gnmi/mixed-schema.md
dplore: I added in response to several comments requesting we expand the scope of this PR to also clarify mixing vendor yang models.  I will remove.

rpc/gnmi/mixed-schema.md
dplore: Will do

rpc/gnmi/mixed-schema.md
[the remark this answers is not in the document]
jsterne: That would put order/timing dependencies back into the architecture. The IP address leaf in the OC model and the IP address leaf in the native model are two different elements in two different parts of the data tree with two different namespaces. The s/w layers below the management layer need decide how to combine those 2 leafs (native win, error if conflict, etc) and it can't really be order based. Consider a reboot scenario where both leafs were configured by the client. We need either native to take precedence or to error on a conflict to have behavior that is deterministic and repeatable across reboots and normal operations.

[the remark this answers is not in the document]
earies: Agree w/ @jsterne - we seem to be going in circles here I feel... the main point here being that an underlying object always has a specified internal order of precedence - not influenced by way of any API order of operations no matter if combined in the same payload/request or separate.

rpc/gnmi/mixed-schema.md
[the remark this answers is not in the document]
earies: It's actually simpler than it looks in the previous text.  Let's put aside the concept of `cli_set_val` and `xml_val` for now since I wanted to focus on how an `origin` and `encoding` work in tandem so we have a better contract and understanding of what origin is meant to entail

If we define an `origin` as a schema-set or tree - we have a clear understanding of what this means - it is a **target tree**.  The current method of `cli` does not mean this - it means a blob of ascii text (tied at the hip to the encoding + which by description means _only_ ASCII - no utf8 which can also be supported) that is left up to interpretation and implementation as to what schema trees can be represented.  If by chance a vendor CLI exposes 3rd party modeling, it could entail this for some and not others...

The reason for the suggested change from `cli` -> `native` here is to accommodate above and also bring this back to what the `openconfig` origin already is.  You now have a nice divide of the tree + encoding.

The reason for the proposed addition of `cli_val` is that this too would be a string and still agreed upon out of band as to "which style" of CLI is the supported syntax.  In the case of JUNOS, this is where we would prefer the `{}` curly brace variant.  Other representations could be specified by precise `TypedValue`'s to further define the contract.

Similarly - in JUNOS we have had concepts for years at the API layer to specify encodings and CLI style variants.

https://www.juniper.net/documentation/us/en/software/junos/junos-xml-protocol/topics/ref/tag/junos-xml-protocol-load-configuration.html

The looser the API or if we are overloading fields, origins, etc... the more issues on interpretation we are going to have and this has already been occurring in many ways to date

rpc/gnmi/mixed-schema.md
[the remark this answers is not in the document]
earies: This imo goes back to proper definition and divide of origins and encodings to resolve this one.

`Get` is problematic from a unary RPC perspective among a few other artifacts I feel but I'll leave that for another discussion.

If we stick w/ `cli` and allow for this in `Subscribe` to circumvent the scaling cases, it would only ever be for the configuration (e.g. `show config`) - why?  Well because origin `cli` is really only focused on config - I will discount the implementation mentioned in another comment here that is leveraging `Get` for packing "show commands" but this would open up an entirely other class of expectations, overloading usage, etc.. that I think is best served via other means.

If we draw this back to an origin of `native` - `Subscribe` would already be intended to be supported for native in order to request modeled config/state (mostly PROTO encoding in reality) - if the encoding is of say the new TypedValue `cli_val` then you only receive config.  As far as the path requested, this is where this would need to be more well defined for this encoding type whether always for the full configuration or a subset by way of a negotiated PathElem "filter" (Since to date we have been mostly focused on XPath representation vs. CLI command arguments)

vcheny:
> Is there a simplification that we can write here that meets the same requirement whilst not having to deal with many edge cases that seem irrelevant?
> 
> 1. We support only `replace` for origin `cli`.
> 2. `replace` with origin `cli` MUST be first in the `SetRequest`.
> 3. Application of another schema -- OpenConfig/vendor native models is then applied on top of this new candidate that was formed by the replace of the CLI. (i.e., replace `cli` means "replace everything on the device even if it's not in this tree").
> 
> This gives us a much simpler solution - that makes not just the `origin: "cli"` specific but also the semantics of a full replace.

As discussed with @robshakir offline, it may worth clarifying how we can support 
1. The whole existing config including OC and VC can be replaced.
2. The non-overlapped VC config  will not be overwritten by OC.
3. Handling VC config that is overlapped with OC.

Looks like `replace cli` can achieve 1. i.e. it will remove everything on the device including VC and OC.
And 2 and 3 can be achieved by only allowing OC `update` when `replace cli` is present. This would also mean OC will overwrite VC if they are overlapped.

dplore:
I will close this PR and make a new one with a more narrow and simple use case like @robshakir suggests.  Our goal is to define one behavior to meet our operational need for a full config replace containing CLI and OpenConfig.    In addition, now that we have learned from the comments and from some internal exploration of concrete configuration scenarios we can provide some concrete examples of interaction between CLI and OC.

jsterne:
> 

In previous calls we were never leaning towards OC overwriting VC. It was always VC overwriting OC (and also some discussion of potentially considering an error if there is overlap).

dplore:
https://github.com/openconfig/reference/pull/188 was introduced to replace this PR

PR #176, opened 2022-11-18
Adjust gNMI SetRequest Update handling for list keys + value payloads

earies:
Clarify that an Update to a list member targeted in the Path cannot include
those list keys in the value payload either as a direct descendant or ultimate
leafref'd node.

Since you cannot modify the list keys on a targeted Path, disallow the contents
altogether in the payload.  At best, the same values _could_ be specified but
since the Path is unprefixed while `JSON_IETF` encoding currently remains
prefixed, an association needs to be drawn from a schema set perspective.

In addition, this reduces pointless and redundant bytes on the wire.

### DIFF
--- a/rpc/gnmi/gnmi-specification.md
+++ b/rpc/gnmi/gnmi-specification.md
@@ -4,10 +4,10 @@
 Paul Borman, Marcus Hines, Carl Lebsack, Chris Morrow, Anees Shaikh, Rob Shakir
 
 **Date:**
-July 7, 2022
+Nov 18, 2022
 
 **Version:**
-0.8.1
+0.9.0
 
 # Table of Contents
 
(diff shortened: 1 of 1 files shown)


earies:
@gcsl - PTAL - this just tightens up the protocol around disallowing list keys to be included in the value payload since they are immutable in this context

earies:
@dplore - looking for review here.  This has been a long standing PR - it looks like some protocol needs to be put in place across various repos to ensure distribution of review - Thx

rpc/gnmi/gnmi-specification.md
```diff
@@ -1283,8 +1291,8 @@ root +
 In this case, nodes `k1` and `k2` are standalone nodes within the schema, but
 also correspond to attribute values for the node "`f`". In this case, an update
 or replace message specifying a path of `/a/f[k1=10][k2=20]` setting the value
-of `k1` to 100 MUST be considered erroneous, and a status code of
-`InvalidArgument (3)` specified.
+of `k1` to either the same value (10) or an alternate value (100) MUST be
```
dplore, line 1294: To support idempotent behavior, I believe setting a node which is also a list key to the same value it already has should be permitted.  Only changing the node's value must be an error.

rpc/gnmi/gnmi-specification.md
```diff
@@ -1247,19 +1247,27 @@ root +
...
-    code of`InvalidArgument (3)` specified.
+  an element - e.g., `/a/f[k1=10][k2=20]`- and a replace or update
+  operation's path specifies  a subset of the attributes (e.g., `/a/f[k1=10]`)
+  then this MUST be considered an error by the target system - and an status
+  code of` InvalidArgument (3)` specified.
+- In the case that key values are specified both as attributes of a node in the
+  path, and as their own elements within the value, then this MUST be considered
+  an error by the target system - and a status code of `InvalidArgument (3)`
+  specified.  Targeting a list member with update or replace operations that
+  modify it's own keys in either conflicting or identical fashion is not
```
dplore, line 1263: To support idempotent behavior, identical update/replace should be permitted.  Agreed that conflicting node / list key values must return an error.  I think that is already covered by the last bullet point in this stanza,

dplore (commented):
There is a behavioral change here regarding support for idempotent behavior that I don't think we want to make.  See comments.

PR #179, opened 2022-12-28
Define special cli and OpenConfig setrequest requirements

dplore:
Clarify use of a gNMI `SetRequest` containing both `origin` `cli` and `origin` `openconfig`

Resolves https://github.com/openconfig/featureprofiles/issues/740

### DIFF
--- a/rpc/gnmi/mixed-schema.md
+++ b/rpc/gnmi/mixed-schema.md
@@ -1,10 +1,11 @@
 # Mixing Schemas in gNMI
-**Contributors**: aashaikh, hines, robjs, csl  
-October 2016, Updated May 2018  
+
+**Contributors**: aashaikh, hines, robjs, csl, dloher
+October 2016, Updated December 2022  
 *Implementation Status*: Merged
 
-
 ## Problem Statement
+
 Today's network devices support multiple schema representations of their
 underlying configuration and telemetry databases. Particularly, devices
 typically support:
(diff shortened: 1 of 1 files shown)


rpc/gnmi/mixed-schema.md
```diff
@@ -108,50 +109,187 @@ if all transactions succeed should the `SetResponse` indicate success. If any
...
+### SetRequest behavior
+
+A `SetRequest` message using the `replace` operation for both `cli` and `openconfig` origins SHOULD be performed as follows:
 
-### Example: OpenConfig and CLI Data
+1. `origin` `cli` SHOULD appear first.  
+2. The entire contents of the `ascii` encoded `cli` data SHOULD appear in a single update.
+3. The `replace` operation for origin `cli` SHOULD be interpeted as a full configuration replacement.  In other words, the target device SHOULD erase the `cli` configuration and replace it with the contents of the `cli` update message.
+4. `origin` `openconfig` should appear second.  
+5. The target device should merge the `cli` and `openconfig` and apply the merged configuration as a single replace operation, replacing the combined `cli` and `openconfig` configuration of the device.
```
aredmon8551, line 131: Sorry in advance - what is the expected behavior in the event of an overlap between them? :)

Advance to step 6?

dplore: Yes step 6, the gNMI server SHOULD respond with an error.   Note in step 6 the reference to transaction behavior.  If the configuration cannot be fully applied, the entire configuration change MUST be rolled back / not applied.

rpc/gnmi/mixed-schema.md
```diff
@@ -108,50 +109,187 @@ if all transactions succeed should the `SetResponse` indicate success. If any
...
+3. The desire to use a full configuration replacement strategy.
+
+To support these use cases a special meaning to a setRequest with replace origin `cli` and replace origin `openconfig` is defined.
+
+### SetRequest behavior
+
+A `SetRequest` message using the `replace` operation for both `cli` and `openconfig` origins SHOULD be performed as follows:
 
-### Example: OpenConfig and CLI Data
+1. `origin` `cli` SHOULD appear first.
```
sebsapa, line 127: what does "appear" mean here?  don't think it means that the client must provide it first, as your example provided does not do that. Perhaps a better term would be "applied"?

Also, the word should/SHOULD is used in the steps where I think the intention is to say MUST.

rpc/gnmi/mixed-schema.md
```diff
@@ -108,50 +109,187 @@ if all transactions succeed should the `SetResponse` indicate success. If any
...
+
+To support these use cases a special meaning to a setRequest with replace origin `cli` and replace origin `openconfig` is defined.
+
+### SetRequest behavior
+
+A `SetRequest` message using the `replace` operation for both `cli` and `openconfig` origins SHOULD be performed as follows:
 
-### Example: OpenConfig and CLI Data
+1. `origin` `cli` SHOULD appear first.  
+2. The entire contents of the `ascii` encoded `cli` data SHOULD appear in a single update.
```
earies, line 128: If these specs are going to try to borrow from RFC2119 keyword terminology then the use of the wording `SHOULD`  everywhere unfortunately defines items that are preferable but not necessary in order to claim compliance.  The end result will be different implementations which is not desired here.

The less strict specs, APIs and data-models are, then divergence can be expected

rpc/gnmi/mixed-schema.md
```diff
@@ -108,50 +109,187 @@ if all transactions succeed should the `SetResponse` indicate success. If any
 of the transactions fail, the contents of all origins MUST be rolled back, and
 an error status returned upon responding to the `Set` RPC.
 
-## Example
+## Special considerations for a setRequest with openconfig and cli origins
+
+### Use cases
+The operational use cases for a setRequest `replace` operation containing both OC and CLI origins include:
```
earies, line 115: As mentioned in the other PRs, we are talking about "mixed schema" here but keep concentrating on CLI only as the native encoding/style

Again this use-case should cover and reference "native" vs. strict CLI usage - this is driving a single use-case that I'm going to venture to bet that other folks that have adopted YANG modeling and other NBI (e.g. NETCONF) are _not_ sending CLI-syntax payloads to network-elements any longer.  The concentration on CLI here leaves a gap in the specification for the cases where we are truly talking about mixed schemas that could be encoded in JSON_IETF or XML already

rpc/gnmi/mixed-schema.md
```diff
@@ -108,50 +109,187 @@ if all transactions succeed should the `SetResponse` indicate success. If any
...
+To support these use cases a special meaning to a setRequest with replace origin `cli` and replace origin `openconfig` is defined.
+
+### SetRequest behavior
+
+A `SetRequest` message using the `replace` operation for both `cli` and `openconfig` origins SHOULD be performed as follows:
 
-### Example: OpenConfig and CLI Data
+1. `origin` `cli` SHOULD appear first.  
+2. The entire contents of the `ascii` encoded `cli` data SHOULD appear in a single update.
+3. The `replace` operation for origin `cli` SHOULD be interpeted as a full configuration replacement.  In other words, the target device SHOULD erase the `cli` configuration and replace it with the contents of the `cli` update message.
```
earies, line 129: What about when multiple schema sets are also represented as CLI syntax?  We seem to keep mixing the schema sets with the presentation layer here

rpc/gnmi/mixed-schema.md
jsterne: In some previous discussions we talked about two potential behaviors when there is overlap:
1) native model data takes precedence (regardless of the order of native origin & OC origin in set request(s))
or
2) error 
Should we consider allowing either of those behaviors?

rpc/gnmi/mixed-schema.md
```diff
@@ -108,50 +109,187 @@ if all transactions succeed should the `SetResponse` indicate success. If any
...
   >
 >
 replace: <
   path: <
-    origin: "cli"
+    origin: "openconfig"
   >
   val: <
-    ascii_val: "router bgp 15169"
+    json_val: `
```
earies, line 157: Should be `json_ietf_val`  - there is no protocol for `json_val` here

rpc/gnmi/mixed-schema.md
jsterne: There are also 2 (maybe more) cases to consider for overlap:
1) The OC and the Native leaf are set to different values
2) The OC and the Native leaf are both explicitly set, but with the same value. 
(it may be simpler to have the same behavior for both of those situations?).
We may also need to define behavior around "default" values (i.e. if Native has a default for a leaf, and the set contains an explicit value for the equivalent leaf, is that "overlap" ?).

rpc/gnmi/mixed-schema.md
hellt: @sebsapa I think it means exactly that - the client should form a Set request in such a way that Updates with origin=cli appear first in the list of Updates.

In the provided example I see the first replace message is specified with `origin=cli` and the 2nd one with `openconfig`

rpc/gnmi/mixed-schema.md
hellt: yes, I understand the use case here with the goal of performing a full-config-replace and providing a CLI blob that is guaranteed to have 100% config coverage.

But this indeed confronts the requirement to keep changes scoped to a certain origin only:

image

I think a form of a deviation statement needs to be given.

rpc/gnmi/mixed-schema.md
hellt: @dplore I wonder if by using SHOULD term here we allow NOSes that treat Openconfig schema as a translation layer on top of native to **not** error?  

In SR Linux, the merge between CLI and Openconfig can be done for overlapping fields -> data that is applied on top will rewrite the existing data. In the case of the ordering used in this case - (1st)CLI 2(nd)OC - data provided with OC schema will overwrite the overlapping data that was initially pushed with CLI.

As I understand, this whole case of overlapping schemas is a guarding rail, as operationally such a combination in a single set request doesn't make much sense.

rpc/gnmi/mixed-schema.md
hellt: @earies, why `json_val` is not allowed anymore?

rpc/gnmi/mixed-schema.md
```diff
@@ -108,50 +109,187 @@ if all transactions succeed should the `SetResponse` indicate success. If any
...
+
+## Examples
+
+### Example: OpenConfig and CLI origin `replace` operation
 
 If a client wishes to replace both OpenConfig and CLI-modelled data
 concurrently, it can send the following `SetRequest`:
 
-```
+```json
```
hellt, line 141: to have a better suitable highlighting
```suggestion
```pbtxt
```

rpc/gnmi/mixed-schema.md
dplore: Agreed, we will introduce a new operation to clarify this.

dplore:
Thank you for all the comments.  I will submit another PR which will replace this one.  It will introduce a new gNMI operation to help clarify the goal of merging CLI and OC together in one replace operation.  My target for publishing that is tomorrow, Feb 3rd, 2023.

hellt:
@dplore are there any news on when the new changeset is going to be published for a discussion?

dplore:
> @dplore are there any news on when the new changeset is going to be published for a discussion?

Yes, a major re-write of this is now complete and being incrementally shared.  A public release will be in March.

PR #187, opened 2023-04-18
Clarify that calling replace on a list element with `{}` is invalid.

wenovus:
The following replace operation should be invalid since it is really a deletion operation.

Alternatively we allow treating it as a delete, but from the current spec it looks like it would be discouraged and I know of a vendor implementation that also errors out in this case.
```
        replace:  {
          path:  {
            elem:  {
              name:  "network-instances"
            }
            elem:  {
              name:  "network-instance"
              key:  {
                key:  "name"
                value:  "RED"
              }
            }
          }
          val:  {
            json_ietf_val:  "{}"
          }
        }
```


dplore (approved):
LGTM

PR #188, opened 2023-05-25
Add union_replace specification

dplore:
The union_replace specification defines how a new gNMI operation is performed which joins two origin datasets together and replaces the currently running configuration.  

This document is a cut and paste of the [gNMI union_replace proposal](https://docs.google.com/document/d/1tuhI_0TymYBGkyEXsJhfdPbCV2UFvUgGh1bUA9eRKaw/edit#heading=h.x9snb54sjlu9) previously shared for OC Community and Operator review. However, there are several clarifications added based on feedback from the community collected into this markdown format. The most important are in the sections pertaining to handling of default values.  Please carefully consider the default handling in case they don't meet your expectations or are unclear.  

Also see the [gNMI proto update](https://github.com/openconfig/gnmi/commit/5473f2ef722ee45c3f26eee3f4a44a7d827e3575)


robshakir (changes requested):
i think this cannot solely be a separate document, but it is meaty enough in its content to warrant one.

can we make changes to the specification to include the high-level operation of `union_replace` with references to this extension document where more clarity is required please?

i'll review the document here (i presume its similar to the one that was previously authored) after we agree here.

dplore:
> i think this cannot solely be a separate document, but it is meaty enough in its content to warrant one.
> 
> can we make changes to the specification to include the high-level operation of `union_replace` with references to this extension document where more clarity is required please?
> 
> i'll review the document here (i presume its similar to the one that was previously authored) after we agree here.

Agreed, the gNMI specification should be updated, I will add updates for it to this PR.

This doc is a cut and paste of the [gNMI union_replace proposal](https://docs.google.com/document/d/1tuhI_0TymYBGkyEXsJhfdPbCV2UFvUgGh1bUA9eRKaw/edit#heading=h.x9snb54sjlu9) shared for OC Community and Operator review.  However, there are several  clarifications added based on feedback from the community collected into this markdown format.  The most important are in the sections pertaining to handling of default values.  Please carefully consider the default handling in case they don't meet your expectations or are unclear.

hellt:
Shall it be somehow synced with gnmi.proto changes and versions bumped synchronously in accordance?

dplore:
> Shall it be somehow synced with gnmi.proto changes and versions bumped synchronously in accordance?

Yes, I am intending to upgrade gnmi.proto to v 0.10.0 as well.

jsterne:
The merge algorithm mentions to include "factory defaults" in 5.2.2 and 5.3.1. Some implementations have the concept of a "factory default" that is loaded if there is no saved config file (i.e. no startup datastore), and that is different than an "empty config" (i.e. delete at root). The factory default may contain, for example, a default user called "admin" or an event log that can be removed/deleted if desired. So I'm not sure if those items (e.g. admin user) should really be included back into the config during a union_replace.

rpc/gnmi/gnmi-union_replace.md
```diff
@@ -0,0 +1,327 @@
...
+
+An abstract set of steps a target may use to achieve `union_replace` behavior are:
+
+1. If using CLI origin, create a candidate configuration using factory defaults within the applicable namespace.  Factory defaults means a configuration with only the default values set by the NOS.  For example a NOS factory default configuration may assert that all ports, except a single management port are shutdown. The NOS may also assert that a DHCP client is enabled on the single management port and that protocols like BGP and ISIS are not instantiated and therefore are also not enabled).  If using another native origin, create a candidate configuration using the current running configuration.
+
+1. Union the native data with the OC paths data and apply this to the candidate  configuration.  The union behavior is described in 5.3 Union behavior options.
```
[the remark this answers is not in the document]
jsterne: I think more clarification is still needed on this.

If in step 1 the intention is that there the possibility of explicit config existing in the running datastore (i.e. "show configuration" returns something, not just a completely blank/empty response), and that config includes statements that are deletable/removable by a client (e.g. a user called "admin"), then the new step 2 proposed would delete all those deletable lines of config. In that case I'm not sure the point of step 1. Anything would just be removed by step 2.

But if the intention after step 1 is that there is no explicit config (i.e. "show configuration" returns absolutely nothing), just an empty config where the NE is using some default values, then we should clarify that in step 1.

jsterne:
A more detailed example of what I mean by "factory default" (and contrasting that with an "empty" configuration):

1) When an network element (NE, e.g. router) boots up, it looks for a previously saved configuration (call this the "startup datastore").
2) If no startup datastore is found, then the NE populates the running datastore with a set of "factory default" config statements. These config statements are indistinguishable from the same config being explicitly set by a client. The config statements appear in the output of "show configuration" (or whatever command the NE uses to view the contents of the running datastore, e.g. gNMI Get). The config statements can be deleted like any other config.
3) An example of some potential factory default config is a user called "admin" or an event log "99".
4) Some operators, for security reasons, don't want to have the factory default "admin" user or event log 99. So they delete them.
5) In their master config up in their client, the operator doesn't have any user called "admin"
6) When the operator does a union_replace they don't want to end up with any user called "admin" in the running config (it isn't up in their master config)

The new proposal by Darren ("and replace" instead of "and apply this to") does achieve this, but then I don't understand the point of step 1. Why not just stay the candidate config starts completely empty ? (note - empty means there are lots of defaults being used by the NE, but defaults aren't explicit config, i.e. don't show up in "show configuration").

jsterne:
Editorial comment: the sections skip from 5.3.3 to 5.3.6.

jsterne:
Section "5.3.3.2 Overlapping values in CLI and OC" describes the two options for resolving issues. Why not allow the same two options for NY and OC?  (5.3.6)?

robshakir (changes requested):

rpc/gnmi/gnmi-union_replace.md
```diff
@@ -0,0 +1,327 @@
+# union_replace gNMI method
```
robshakir, line 1: (nit) can you 80ch wrap this file please?

rpc/gnmi/gnmi-union_replace.md
```diff
@@ -0,0 +1,327 @@
...
+  - [6 Examples](#6-examples)
+    - [6.1 Examples of expected configuration work flows](#61-examples-of-expected-configuration-work-flows)
+    - [6.3 union\_replace example](#63-union_replace-example)
+  - [7. gNSI, bootz and gNMI interoperation](#7-gnsi-bootz-and-gnmi-interoperation)
+  - [8. References](#8-references)
+
+## 1. Objective
+
+There is an operational goal to ensure a network operating system’s (NOS) complete operating configuration state is matched to the intended configuration state.  A simple way to guarantee this is by generating the entire configuration for the device and replacing whatever configuration the device currently has.  This declarative specification of the configuration is colloquially referred to as ‘full config replace’.
+Although the concept is simple enough, there are portions of the configuration which may not be modeled in OpenConfig.
```
robshakir, line 51: Suggest to add something here that states something like, "gNMI already supports this concept through the definition of multiple schemas, see https://github.com/openconfig/reference/blob/master/rpc/gnmi/mixed-schema.md".

We should consider what our approach for this document's coexistence with the previous document is -- the `origin` field's meaning is established in the mixed schema document. Maybe removing the example from mixed schema (or making it simpler to include non-overlapping information) and then referencing this document for more complex cases is something that would ensure usability for documentation readers.

rpc/gnmi/gnmi-union_replace.md
```diff
@@ -0,0 +1,327 @@
...
+
+### 2.8 bootz
+
+[bootz](https://github.com/openconfig/bootz) describes structured data and API for taking a device from factory defaults to a ready for production state.  bootz is intended to replace secure zero touch provisioning (sZTP) solutions.
+
+Bootz defines partitions or namespaces for NOS configuration items and an order for applying these namespaces to the NOS.  This is described briefly later in this document and in detail at the [bootz](https://github.com/openconfig/bootz) repository.
+
+### 2.9 gNSI
+
+gNSI requires modeled config paths to be disabled or restricted from OC and NY paths.  If these gNSI controlled paths are attempted to be set using a gNMI `SetRequest`, the `SetRequest` must be rejected with an error.
```
robshakir, line 91: It's not clear to me what this first sentence means. gNSI allows a policy to be set that can control access to these paths, but doesn't require disabled or restricted by default AFAICS. Can you clarify?

rpc/gnmi/gnmi-union_replace.md
```diff
@@ -0,0 +1,327 @@
...
+A configuration item is any attribute that may be assigned some value that is then stored in a NOS that affects the behavior of the NOS or its supported services.
+
+### 2.11 Full Device Configuration
+
+This term must be avoided. Because each NOS varies in which configuration items are part of its native configuration, this term must not be used to describe the intended configuration of a device.
+Examples of the variation in what is included in a ‘full device configuration’  include NOS specific knobs, hardware specific configuration, kernel boot information and security configuration items.
+
+### 2.12 Dynamic configuration
+
+Configuration affected by a gNMI SetRequest is defined to be Dynamic Configuration.  The term ‘Dynamic’ is used because gNMI configuration is expected to be modified from time to time while the NOS is operationally in service and forwarding traffic.
```
robshakir, line 104: (nit) SetRequest should be `SetRequest` for formatting, here and throughout.

rpc/gnmi/gnmi-union_replace.md
```diff
@@ -0,0 +1,327 @@
...
+
+### 3.2 What origins are to be mixed?
+
+There are two key use cases for mixing configuration schemas:
+
+A network operator is transitioning from existing CLI-based configuration generation to OpenConfig, and wants to achieve this incrementally. This, by definition, requires mixing CLI text with OpenConfig modeled output.
+A network operator wants to produce entirely modeled configuration, but needs a schema other than OpenConfig to express some features, as discussed in the previous section.
+
+Thus, mixed schema must handle two combinations:
+OC and CLI
```
robshakir, line 135: (nit) Make these bullet points.

rpc/gnmi/gnmi-union_replace.md
```diff
@@ -0,0 +1,327 @@
...
+OC and NY
+
+Note, it is assumed that both the native schema and CLI model the entire configuration surface area of the device. There is no requirement to support OC, CLI & NY at the same time.
```
robshakir, line 138: `s/native schema/native YANG/`?

rpc/gnmi/gnmi-union_replace.md
```diff
@@ -0,0 +1,327 @@
...
+
+Furthermore note that these origins contain contents that overlap – which is the motivation for the union_replace operation. Where origins are entirely disjoint, they can be manipulated independently (whether in the same `SetRequest` or different ones) since there is no interdependency between them.
```
robshakir, line 140: Here we might want to link back to the simpler mixed schema document.

rpc/gnmi/gnmi-union_replace.md
```diff
@@ -0,0 +1,327 @@
...
+
+## 4. Device behavior for single-origin SetRequests
+
+### 4.1 Replace operation for Origin CLI  behavior
+
+We first define a scenario to perform a replace of configuration items with only CLI origin (not mixed origins).  Origin CLI does not support addressing a path and is therefore interpreted as the empty path, `/` (ie: the “root” path) which implies the entire contents of the CLI origin.  If bootz or gNSI are enabled then the gNMI SetRequest MUST not replace any bootz or gNSI managed configuration items.  In other words, the CLI origin replace only replaces the configuration items in the dynamic config namespace.
+
+When performing a gNMI SetRequest `replace` operation, the entire configuration that the CLI  manages (subject to the namespace rules) should be ‘erased’ and replaced with the contents of the CLI ASCII blob value.  It is acknowledged that the entire CLI configuration may contain  configuration items which overlap with other origins.  CLI replace should only erase configuration that is writable by CLI.  For example, certificates should not be erased by a setRequest in a scenario where installation of certificates is not managed by CLI.
```
robshakir, line 148:
```
s/setRequest/`SetRequest`/
```

rpc/gnmi/gnmi-union_replace.md
```diff
@@ -0,0 +1,327 @@
...
+
+A gNMI `SetRequest` `replace` operation using origin CLI must use only one  `replace` operation which contains the entire contents of the desired CLI configuration.  One or more `update` operations may follow the `replace` operation. The contents of the `update` are appended to the contents of that specified in the `replace` operation.
+
+Note, in a gNMI SetRequest the `delete`, `replace` and `update` operations are performed in order as prescribed by the gNMI specification.  In origin CLI, the delete operation is unsupported.
+
+### 4.2 Replace operation for path based Native Origin behavior
+
+A `replace` operation for path based native origins should be performed in the same way as OpenConfig as defined in gNMI Specification section 3.4.
+
+## 5 Device behavior for mixed native/CLI OC origin  SetRequests
```
robshakir, line 160:
```suggestion
## 5 Device behavior for mixed native/CLI OC origin `Set` RPCs
```

rpc/gnmi/gnmi-union_replace.md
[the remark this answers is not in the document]
robshakir: If a line of config is deletable but not default, why is it returned by 1? As I read it, anything that is a default, whether returned by some legacy CLI method ;-) or not should be in the candidate that is to be applied.

robshakir, line 181: Update the 5.3 to be a link to the relevant section.

rpc/gnmi/gnmi-union_replace.md
```diff
@@ -0,0 +1,327 @@
...
+
+An abstract set of steps a target may use to achieve `union_replace` behavior are:
+
+1. If using CLI origin, create a candidate configuration using factory defaults within the applicable namespace.  Factory defaults means a configuration with only the default values set by the NOS.  For example a NOS factory default configuration may assert that all ports, except a single management port are shutdown. The NOS may also assert that a DHCP client is enabled on the single management port and that protocols like BGP and ISIS are not instantiated and therefore are also not enabled).  If using another native origin, create a candidate configuration using the current running configuration.
+
+1. Union the native data with the OC paths data and apply this to the candidate  configuration.  The union behavior is described in 5.3 Union behavior options.
+
+1. If an error is experienced when performing the union, the target responds to the RPC with a canonical error code of `INVALID_ARGUMENT` and SHOULD populate the error details with sufficient information for the union operation’s failure to be debugged.
+
+1. Replace the running configuration with the candidate configuration
```
robshakir, line 185:
```suggestion
1. Replace the running configuration with the candidate configuration.
```

rpc/gnmi/gnmi-union_replace.md
```diff
@@ -0,0 +1,327 @@
...
+
+1. If an error occurs when performing the configuration replacement the existing gNMI specification 3.4.3 applies.  The target rolls back all configuration changes.  The target responds to the RPC with a canonical error code of `INTERNAL` and SHOULD populate the error details with sufficient information for the replacement operation’s failure to be debugged.
+
+### 5.3 Union behavior options
+
+This section describes how the union between native and OC should be performed.  Different scenarios are described for handling CLI vs. Native origin union with OpenConfig.
+In addition, there are two options for handling errors in the union process. A NOS must assert which error handling option is implemented in it's documentation.
```
robshakir, line 192:
```suggestion
In addition, there are two options for handling errors in the union process. A NOS must assert which error handling option is implemented in its documentation.
```

rpc/gnmi/gnmi-union_replace.md
```diff
@@ -0,0 +1,327 @@
...
+2. Process OC paths, replacing the relevant subtree(s) in the candidate configuration.  This includes OC default values.
+3. Process native paths, replacing the relevant subtree(s) in the candidate configuration.  Generate an error if there is conflict between OC and Native as defined below.
+
+#### 5.3.3 Resolving issues with union between the origins
+
+##### 5.3.3.1 Default values
+
+The following rules govern how default values are applied when using two origins.
+
+1. Defaults from OC apply at the container level.
```
robshakir, line 220: What does this mean?

rpc/gnmi/gnmi-union_replace.md
```diff
@@ -0,0 +1,327 @@
...
+1. If the same configuration item is explicitly set by both origins to the same value, then that is accepted.
+1. If a configuration item is explicitly set in one origin, but default in the other origin, then the explicit value always takes effect.
+1. If a configuration item is not set and has only a default value in both origins then the OC default must be used.
+1. If a configuration item is explicitly set by both origins to two different values then it should be handled by one of the options in 5.3.5 below.
+
```
robshakir, line 225: I would like some text here that states specifically "if you want to have the default value from a particular origin apply explicitly, the safest thing to do is to explicitly specify that default within that origin's payload".

rpc/gnmi/gnmi-union_replace.md
```diff
@@ -0,0 +1,327 @@
...
+When applying CLI or NY data to the candidate configuration, overwrite any configuration item already present following the order of merge precedence.
+
+#### 5.3.6 Overlapping values in NY and OC
+
+Where a pathed, NY origin configuration item overlaps with OC, the target MUST return an `INVALID_ARGUMENT` error.  It is up to the client to provide the native schema and OC which are not in conflict.
+
+#### 5.3.7 Example of resolving default values
+
+Here are two examples of when to use OC or CLI defaults in a setRequest.
```
robshakir, line 244:
```suggestion
Here are two examples of when to use OC or CLI defaults in a `SetRequest`.
```

rpc/gnmi/gnmi-union_replace.md
```diff
@@ -0,0 +1,327 @@
...
+2. at a later time in a later s/w release, feature A becomes accessible via OC
+3. configuration push omits the configuration in CLI and includes it in OC via union_replace.
+4. The expected union_replace behavior is that the feature remains configured and no error is returned.  Abstracted from the client, the 'view' containing the feature is changed from CLI to OC.
+
+### 6.3 union_replace example
+
+```json
+union_replace: {
+  path: {
+    origin: "cli"
```
robshakir, line 270:
```suggestion
    origin: "acme_cli"
```

based on the naming suggestions above?

robshakir (changes requested):

rpc/gnmi/gnmi-union_replace.md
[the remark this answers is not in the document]
jsterne: I'm not positive what's being described in #1 wrt "factory default", but worried that term may cause confusion since it is a standard term in the IETF. Factory default config in the IETF looks & feels like a small set of config that a user/client explicitly added.  In other words, a gNMI Get would return some data (e.g. user "admin", etc).

But that's very different from default values in a schema. For example: let's pretend an interface mtu has a default of 1500 in a YANG schema. If a client never sets that mtu leaf, then reading back the config with a Get would not return any leaf called mtu. But the system would be using an mtu of 1500 down in the app layer. IMO I don't consider the mtu leaf to be in the configuration/candidate (i.e. there is nothing there in the instance data for the mtu).

I'm not clear if #1 is saying a Get would return:
(A) absolutely no data (i.e. default values are being used in the app layer, but there is no config at all in the running DS), or 
(B) the Get would return some data (i.e. like the IETF "factory default" scenario, where maybe a user "admin" might be returned in the get response).  

From what I can tell it must be (A). But then why don't we just say the following for #1:
1. Start with an empty candidate configuration (no instance data, a Get would return an empty data set)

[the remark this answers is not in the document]
jsterne: I'm similarly unsure about the last sentence of bullet 1: _If using another native origin, create a candidate configuration using the current running configuration._

If step 2 replaces the config, then why are we starting with the current running vs an empty candidate? Maybe we'd want to mention that either is equivalent (to clarify there isn't something subtle/special here about starting with the current running).

rpc/gnmi/gnmi-specification.md
```diff
@@ -922,10 +919,14 @@ it desires.
...
-3.4.3](#343-transactions)). The server MUST process deleted paths (within the
-`delete` field of the `SetRequest`), followed by paths to be replaced (within
-the `replace` field), and finally updated paths (within the `update` field). The
-order of the replace and update fields MUST be treated as significant within a
+3.4.3](#343-transactions)). The server MUST process paths in the following order:
+
+1. deleted paths (within the`delete` field of the `SetRequest`)
+2. union_replaced paths (within the `union_replace` field)
+3. replaced paths (within the `replace` field)
+4. updated paths (within the `update` field)
```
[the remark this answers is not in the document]
dplore: I will update to specify union_replace is an exclusive operation, not combined with the others

rpc/gnmi/gnmi-union_replace.md
dplore: I will move the detailed description of 'origin' from the mixed-schema doc to the gnmi specification.  I think the mixed schema doc then would be left only with things that overlap with union_replace and could be deleted.  I'll send another commit with the changes for your view

rpc/gnmi/gnmi-union_replace.md
dplore: I will reference the yang specification for default applications.

rpc/gnmi/gnmi-union_replace.md
dplore: Will add this recommendation.

jsterne: Maybe instead of "specify that default" says "specify that value" ?

rpc/gnmi/gnmi-union_replace.md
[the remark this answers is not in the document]
dplore: I will add a definition for factory default configuration.  It will read something like:
"The configuration that is present on the device when it is powered on for the first time by the user."

When using union_replace, the user is expecting to assert all configuration items needed to ensure the device has the intended state.    If a device factory default configuration includes an admin account and the user wants this removed, the user must include the negation of that user in the union_replace.  

gNMI operations, including union_replace, are expected to be idempotent.  So each and every union_replace operation would include this negation, but it would only cause a change in the 'running' configuration on the first SetRequest.  Subsequent SetRequest would include the negation, but there would be no effect on the 'running' configuration or 'startup' configuration because that admin account was already removed.  

Regarding  native yang (NY) origin the user can choose to replace from the root NY tree or at subtrees.  In CLI, this is not feasible because there is no path structure, so we chose to require a 'factory default' candidate configuration.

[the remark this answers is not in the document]
jsterne: The negation of an admin account (in a declarative config) is simply the absence of it in the combined native+OC union config that was sent in the union_replace. When you say "must include the negation" it makes it sound like you want the instruction "no user admin" somehow in the union_replace (i.e. an imperative statement from the "old days" vs a declarative configuration).

rpc/gnmi/gnmi-union_replace.md
dplore: Done

rpc/gnmi/gnmi-union_replace.md
dplore: Updated with reference to pathz and linked to bootz's definition of interoperation with gNSI.

rpc/gnmi/gnmi-union_replace.md
dplore: done

rpc/gnmi/gnmi-union_replace.md
dplore: Done

rpc/gnmi/gnmi-union_replace.md
dplore: Done

rpc/gnmi/gnmi-union_replace.md
dplore: Got it.  Added this:

As a user of this specification, to ensure the desired value is applied, the
safest thing to do is to explicitly specify that value within the desired
origin's payload.

rpc/gnmi/gnmi-union_replace.md
dplore: Updated

dplore:
Last call for comments.  This will merge on June 10th, 2023.

@rgwilton
@rolandphung
@hellt

rpc/gnmi/gnmi-specification.md
```diff
@@ -922,10 +919,14 @@ it desires.
...
 within it - which are treated as a transaction (see [Section
-3.4.3](#343-transactions)). The server MUST process deleted paths (within the
-`delete` field of the `SetRequest`), followed by paths to be replaced (within
-the `replace` field), and finally updated paths (within the `update` field). The
-order of the replace and update fields MUST be treated as significant within a
+3.4.3](#343-transactions)). The server MUST process paths in the following order:
+
+1. deleted paths (within the`delete` field of the `SetRequest`)
+2. union_replaced paths (within the `union_replace` field)
+3. replaced paths (within the `replace` field)
```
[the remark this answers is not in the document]
dplore: Elaborated a little more

dplore:
@jsterne I have responded to all your comments.  PTAL at the latest revision.

rpc/gnmi/gnmi-specification.md
```diff
@@ -1659,6 +1758,9 @@ limitations under the License
 
 # 7 Revision History
 
+* v0.10.0: May 25, 2023
+  * Add union_replace operation.  Sync revision to gNMI proto revision.
+
```
hellt, line 1763: @dplore I am curious, should this PR not be merged before the gnmi.proto is merged? The text here says "sync revision with gnmi proto", but gnmi proto is not even in a PR state"

dplore: gnmi proto v0.10.0 is now published at: https://github.com/openconfig/gnmi/commit/5473f2ef722ee45c3f26eee3f4a44a7d827e3575#diff-49fc5824be3cfb8d64321cc64142f0568a3039a55da00d2aaed8fd766369854c

rpc/gnmi/gnmi-specification.md
```diff
@@ -1,76 +1,83 @@
...
-Paul Borman, Marcus Hines, Carl Lebsack, Chris Morrow, Anees Shaikh, Rob Shakir, Wen Bo Li
+Paul Borman, Marcus Hines, Carl Lebsack, Chris Morrow, Anees Shaikh, Rob Shakir, Wen Bo Li, Darren Loher
 
 **Date:**
-April 18, 2023
+May 25, 2023
 
 **Version:**
-0.8.3
+0.10.0
```
robshakir, line 10: Is there anything that changed in 0.9.0 that we didn't cover in this update? Can we just file an issue if there is so that we don't lose track of it.

rpc/gnmi/gnmi-specification.md
```diff
@@ -688,20 +690,101 @@ the catalog entry.
...
+constrained to the specified origin. Particularly:
+
+* `replace` operations MUST only replace the contents of the specified `origin`
+  at the specified path. Origins that are not specified within the `SetRequest`
+  MUST NOT have their contents replaced. In order for a `replace` operation to
+  replace any contents of an `origin` it must be explicitly specified in the
+  `SetRequest`.
+* `delete` operations MUST delete only the contents at the specified path within
+  the specified `origin`. To delete contents from multiple origins, a client
+  MUST specify multiple paths within the `delete` of the `SetRequest`.
```
robshakir, line 771: I think we said here that we would note something like:

> These rules apply where origins represent data that doesn't overlap, in some cases (e.g., CLI and OpenConfig) origins may reflect different 'views' on the same data, and thus their interaction is more complex. See <reference to union replace> for more details, and behavioural specifications for such origins.

rpc/gnmi/gnmi-specification.md
```diff
@@ -688,20 +690,101 @@ the catalog entry.
...
+of the transactions fail, the contents of all origins MUST be rolled back, and
+an error status returned upon responding to the `Set` RPC.
+
+## 2.8 Extensions to gNMI
 
 Each top-level RPC message (e.g., `SubscribeRequest` and `SubscribeResponse` for
 the `Subscribe` RPC) defines an `extensions` field which can be used to carry
 additional parameters for a gNMI RPC.
-[[GNMI-EXT]](https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-extensions.md)
+[[GNMI-EXT]](<https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-extensions.md>)
```
robshakir, line 787: This doesn't look correct.

rpc/gnmi/gnmi-specification.md
```diff
@@ -922,10 +1005,15 @@ it desires.
...
-`delete` field of the `SetRequest`), followed by paths to be replaced (within
-the `replace` field), and finally updated paths (within the `update` field). The
-order of the replace and update fields MUST be treated as significant within a
+3.4.3](#343-transactions)). The server MUST process paths in the following order:
+
+1. deleted paths (within the`delete` field of the `SetRequest`)
+2. replaced paths (within the `replace` field)
+3. updated paths (within the `update` field)
+or as a single set of union_replace operations.  That is, union_replace is not
+mixed with other operations in a SetRequest.
```
robshakir, line 1014: This wording isn't that clear to me. I think it's better to just say.

> Where a `SetRequest` contains `delete`, `replace`, and `update` fields, the server MUST process paths in the following order:
> 1. deleted paths...
> 2. replaced paths...
> 3. updated paths...
>
> A `SetRequest` containing `delete`, `replace` and `update` MUST NOT contain `union_replace` operations.

rpc/gnmi/gnmi-specification.md
```diff
@@ -922,10 +1005,15 @@ it desires.
 
 A target receiving a `SetRequest` message processes the operations specified
 within it - which are treated as a transaction (see [Section
-3.4.3](#343-transactions)). The server MUST process deleted paths (within the
-`delete` field of the `SetRequest`), followed by paths to be replaced (within
-the `replace` field), and finally updated paths (within the `update` field). The
-order of the replace and update fields MUST be treated as significant within a
+3.4.3](#343-transactions)). The server MUST process paths in the following order:
+
+1. deleted paths (within the`delete` field of the `SetRequest`)
```
robshakir, line 1010:
```suggestion
1. deleted paths (within the `delete` field of the `SetRequest`)
```

rpc/gnmi/gnmi-specification.md
```diff
@@ -922,10 +1005,15 @@ it desires.
...
+2. replaced paths (within the `replace` field)
+3. updated paths (within the `update` field)
+or as a single set of union_replace operations.  That is, union_replace is not
```
robshakir, line 1013: `union_replace` should be code backticks throughout for consistency.

rpc/gnmi/gnmi-specification.md
```diff
@@ -950,17 +1038,20 @@ the success or failure of the set of operations within the `SetRequest` message
...
-*   `prefix` - specified as per [Section 2.4.1](#241-path-prefixes). The prefix
+* `prefix` - specified as per [Section 2.4.1](#241-path-prefixes). The prefix
     specified is applied to all paths defined within other fields of the
     message.
-*   `delete` - A set of paths, specified as per [Section 2.2.2](#222-paths),
+* `delete` - A set of paths, specified as per [Section 2.2.2](#222-paths),
     which are to be removed from the data tree. A specification of the behavior
     of a delete is defined in [Section 3.4.6](#346-deleting-configuration).
-*   `replace`  - A set of `Update` messages indicating elements of the data tree
+* `union_replace` - A set of `Update` messages specifying elements to union and
```
robshakir, line 1047:
```suggestion
* `union_replace` - A set of `Update` messages specifying elements to union (that is merge) and
```

rpc/gnmi/gnmi-specification.md
robshakir: Here, a forward reference to what `union_replace` is would be useful (just a link to the relevant section).

rpc/gnmi/gnmi-specification.md
```diff
@@ -950,17 +1038,20 @@ the success or failure of the set of operations within the `SetRequest` message
...
+    then replace the data tree.  See the [gNMI union_replace specification](https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-union_replace.md)
```
robshakir, line 1048:
```suggestion
    then replace the data tree.  See the [`union_replace` specification](https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-union_replace.md)
```

rpc/gnmi/gnmi-specification.md
```diff
@@ -1043,19 +1134,24 @@ entries that are not supplied in the value provided in the `SetRequest`. An
 update operation MUST be considered to add new entries to the collection if they
 do not exist.
 
+For `union_replace` operations, the contents contents shall be merged (unioned)
+and then applied as a replacement of the configuration.  See the
+[gNMI union_replace specification](https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-union_replace.md).
```
robshakir, line 1139:
```suggestion
[`union_replace` specification](https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-union_replace.md).
```

rpc/gnmi/gnmi-specification.md
```diff
@@ -1221,21 +1320,21 @@ the status message returned within the RPC response MUST be set to a specific
...
     `InvalidArgument (3)`. In this case, the `message` field of the returned
     status specified SHOULD specify human-readable text indicating that the path
     could not be parsed.
-*   When the operation is an update or replace operation that corresponds to a
-    path that is not valid - `NotFound (5)`. In this case the `message` field of
-    the returned status message SHOULD specify human-readable text indicating
-    the path that was invalid.
-*   When the operation is an update or replace operation that includes an
-    invalid value within the `Update` message specified - `InvalidArgument (3)`.
+* When the operation is an update, replace or union_replace operation that
```
robshakir, line 1332:
```suggestion
* When the operation is an `update`, `replace` or `union_replace` operation that
```

rpc/gnmi/gnmi-specification.md
```diff
@@ -1221,21 +1320,21 @@ the status message returned within the RPC response MUST be set to a specific
...
+    corresponds to a path that is not valid - `NotFound (5)`. In this case the
+    `message` field of the returned status message SHOULD specify human-readable
+    text indicating the path that was invalid.
+* When the operation is an update, replace or union_replace operation that includes
```
robshakir, line 1336:
```suggestion
* When the operation is an `update`, `replace` or `union_replace` operation that includes
```

rpc/gnmi/gnmi-specification.md
```diff
@@ -1659,6 +1758,9 @@ limitations under the License
 
 # 7 Revision History
 
+* v0.10.0: May 25, 2023
+  * Add union_replace operation.  Sync revision to gNMI proto revision.
```
robshakir, line 1762:
```suggestion
  * Add `union_replace` operation.  Sync revision to gNMI proto revision.
```

rpc/gnmi/gnmi-union_replace.md
```diff
@@ -0,0 +1,525 @@
...
+
+### 5.3 Union behavior options
+
+This section describes how the union between native and OC should be performed.
+Different scenarios are described for handling CLI vs. Native origin union with
+OpenConfig. In addition, there are two options for handling errors in the union
+process. A NOS must assert which error handling option is implemented in its
+documentation.
+
+#### 5.3.1 union behavior for CLI and OpenConfig
```
robshakir, line 329:
```suggestion
#### 5.3.1 Union behavior for CLI and OpenConfig
```

rpc/gnmi/gnmi-union_replace.md
```diff
@@ -0,0 +1,525 @@
...
+     - Factory defaults are used because CLI is expected to represent the entire
+      configuration, subject to any partitioning rules of bootz and gNSI.
+2. Process OC paths, replacing the relevant configuration items in the candidate
+   configuration.  This includes OC default values.
+3. Process CLI items merging them to the candidate configuration.  Generate an
+   error if there is conflict between OC and CLI as defined below.  Conflict and
+   resolution is defined in
+   [5.3.3](5.3.3-Resolving-issues-with-union-between-the-origins).
+
+#### 5.3.2 union behavior for Native and OpenConfig
```
robshakir, line 351:
```suggestion
#### 5.3.2 Union behavior for Native and OpenConfig
```

robshakir (changes requested):

rpc/gnmi/gnmi-specification.md
dplore: Version 0.9.0 appears to be all whitespace changes.   Shouldn't have bumped the minor for that.   Ref:
 https://github.com/openconfig/gnmi/commit/d5360e33fc3b22effeaaaf55f345c1f950765bed#diff-49fc5824be3cfb8d64321cc64142f0568a3039a55da00d2aaed8fd766369854c

rpc/gnmi/gnmi-specification.md
dplore: Apparently markdown lint md034 - bare url says there should be <> enclosing a URL.  I haven't noticed that before myself.  The link is working when rendered in github and VSCode.  So I guess I'll leave it?

rpc/gnmi/gnmi-specification.md
dplore: Done

rpc/gnmi/gnmi-specification.md
dplore: Added

rpc/gnmi/gnmi-specification.md
dplore: I think the linter I am using is not so good on this point and the rationale for this not so strong, so I have removed the angle brackets for URL's for consistency across the doc.

robshakir (approved):
LGTM. Thanks for the edits.

hellt:
Can we hold it from merging for a couple of days more?

jsterne:
> @jsterne I have responded to all your comments. PTAL at the latest revision.

Things are looking good for section 5.3.2 (NY+OC) is good (without a path, or a root path, means that everything in step 1 is removed).

It was really the last updates about "negation" that was causing me the most concern. The formatting of section 5.2.2 may be throwing me off (bullet numbering and indentation when I view it here: https://github.com/openconfig/reference/blob/63657555ae3ddddb44474a437d08011b247bf808/rpc/gnmi/gnmi-union_replace.md#532-union-behavior-for-native-and-openconfig)
but we may want to be more clear that this "negation" part only applies to the native CLI part (i.e. is only relevant for CLI+OC).

dplore:
> > @jsterne I have responded to all your comments. PTAL at the latest revision.
> 
> Things are looking good for section 5.3.2 (NY+OC) is good (without a path, or a root path, means that everything in step 1 is removed).
> 
> It was really the last updates about "negation" that was causing me the most concern. The formatting of section 5.2.2 may be throwing me off (bullet numbering and indentation when I view it here: https://github.com/openconfig/reference/blob/63657555ae3ddddb44474a437d08011b247bf808/rpc/gnmi/gnmi-union_replace.md#532-union-behavior-for-native-and-openconfig) but we may want to be more clear that this "negation" part only applies to the native CLI part (i.e. is only relevant for CLI+OC).

I left it open on purpose because I was thinking one could choose to change a factory default using their preferred origin (or available origin in the case where the factory default is not modeled in OC or NY).

PR #190, opened 2023-06-17
Fix numbering scheme in union replace

hellt:
Hi @dplore 

The numbering scheme is not correct, it seems. Let me know if that was not the original intention. 

This PR fixes the numbering to be rendered like this:
image


dplore (approved):

dplore:
Thanks @hellt !

PR #191, opened 2023-06-17
fix union replace combinations

hellt:
@dplore 
Looks like the combinations qualified for mixed origins are not correct?

Sorry if my assumption is incorrect.


dplore (approved):
Thank you for the fix.  Yes, this is the intent.

Also to be explicit about what we aren't including: we don't have an operational use case to mix CLI and NY so we didn't require support for this combination.

PR #196, opened 2023-09-20
Add commit confirmed extension to SetRequest

kidiyoor:


hellt:
Hi @kidiyoor 
is there a proto defining this proposed extension?
We at Nokia created a custom extension for this as well, but if you're working on a registered extension then it makes sense to converge, I believe.

kidiyoor:
@hellt yes, the goal is to converge on the extension. You can find the proto in the proposed md file. Could you share how you r custom extension look like ?

hellt:
@kidiyoor yes, we would like to collaborate on this one. I will have to gather some details internally before pasting our extension proto.

earies:
A few initial comments (will just lump here vs. inline) as we start to breach upon more reinvention of long standing features from other well-defined protocols.

* What is the motivation behind this being a completely separate extension vs. baked directly into gNMI?  I will assume desired behavior for some but not all?
* The behaviors described here are similar yet diverge RFC6241 style confirmed-commit capabilities.  What is the motivation for divergence?  While the feature itself is well understood to transactional models of the current/past implementations, can you elaborate on why the precise behaviors of RFC6241 are insufficient requiring divergence across all implementations that have provided this support for various years? (e.g. rational behind the reinvention to fit a concept into gNMI that was unaccounted for prior)
* Section 2.2.2 states that the during the `confirm` call, a client _should_ send the **same** configuration followed by subsequent logic to compare the configuration to the currently active running configuration (that was previously committed with a `rollback_duration`.  This is not how RFC6241 style confirmation works but rather is suggesting a new feature that imo wastes bytes on the wire and now requires additional logic and processing on the element.  A confirmation in the NETCONF world is just "another commit" that can be performed by any client (either existing session or new) - this subsequent commit could be to confirm or introduce additional changes.  (See `<persist-id>` here as well for the association to client sessions) - This similar logic could be used here to assign an identifier for later confirmation vs. resending an entire payload
* Section 2.2.3 also diverges behaviors from RFC6241.  Why is there a need to prevent another commit/Set whilst the duration timer is in effect?
* Is the plan to rely solely off error-messaging to understand if an element is currently in a `rollback_duration` time window?  This is where I think a state leaf would be helpful (or 2) to indicate if an element is currently in this state (e.g. `commit-rollback=true`) and how much time may be left (e.g. `commit-rollback-timer=27`)

In reality, gNMI Set() abstracts out important distinct transactional operations that the NETCONF world has had for some time.  These operations have been part of operational workflows for some time (whether good or bad) and while this may introduce back a feature from this world, other gating factors for various operators have also been `rollback` as well as any `diff` capabilities in pre-staging environments (which relies off of datastores more and more).  If there is ever consideration of introducing those features back either directly or via extensions, they should also be considered or a best practice guide indicating the reason why these features may never exist for the intentions of gNMI operations.

robshakir:
Ebben -- whilst I'm sympathetic to "some operators might need additional things", all of the OpenConfig work (and I'd say open source contributions in general) are going to be driven by a use case that exists today. If these operators that want to have a wider transactional framework in gNMI (which I'm unsure that I _personally_ am supportive of, for the record), then can we go ahead and discuss that in the context of their use cases -- along with contributions that move this forward please?

I'm not clear that deviations from 6241 are a first-class problem -- these are different protocols with different designs. Implementation decisions that cause operational overhead/problems, and/or cannot be implemented in shipping implementations are more of concern. I do not consider "we baked this in an RFC" as a yardstick for excellent (or even implementable) technical design.

w.r.t the specific question of "why an extension?". Practically, it is extremely complex for us to keep a single specification that describes everything in gNMI. Extensions allow the protocol to be extended in a manner that allows new use cases to be covered that aren't necessarily relevant to all implementations (see master election for another example). On top of that, it allows us to provide subsets of the protocol that can be described in more constrained specifications. Practically, my preference is to extend gNMI _primarily_ via extensions going forward. In some cases, this isn't possible (see `union_replace` as something that would be extremely difficult to add through an extension), so base specification changes are required, but generally, I feel if we can be more constrained that will be more comprehensible to users as the use cases gNMI covers expand.

thx,
r.

dplore:
+1 on extension being the primary way to update the gNMI proto.  The base gNMI proto should be considered a minimum implementation.  With extensions we create additional features that are in addition to the base proto.    The target for this extension is to be a [well known extension](https://www.google.com/url?q=https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-extensions.md%23well-known-and-registered-extensions&sa=D&source=docs&ust=1696456143067839&usg=AOvVaw39JGQtaYAAps3ciITjj57O) 

Regarding the requirements for the `confirm` message, I think we can have a debate about what is efficient and supports the operational use cases.  We should avoid referencing prior art as the only justification.   One should reference operational use cases (#1) , efficiency and  existing implementations.

hellt:
From Nokia perspective I am aiming  to post our proposal and implementation example next week.

Will do this as a PR against the current master branch for an easier review process

rpc/gnmi/gnmi-commit-confirmed.md
```diff
@@ -0,0 +1,58 @@
...
+### 2.2.2 Confirm
+
+During the *confirm* call the client should send the same configuration in the SetRequest spec along with
+the *confirm* field set to true.
+
+If the confirm call is issued with a different configuration, a FAILED_PRECONDITION error is returned.
+
+The confirm call can only be issued if the server is waiting for confirmation. If a SetRequest RPC is received with the
+*confirm* field set to true but the server is not waiting for a confirmation then an FAILED_PRECONDITION error is
+returned.
```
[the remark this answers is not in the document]
kidiyoor: That is correct. It also means we will return error if both rollback_duration is set and confirm is set to true in the initial SetRequest as server is not waiting for a confirmation.

rpc/gnmi/gnmi-commit-confirmed.md
```diff
@@ -0,0 +1,58 @@
...
+
+### 2.2.3 Multiple SetRequest
+
+If a subsequent SetRequest RPC is received with the *rollback_duration* field set, whilst an existing rollback counter
+is running, the server shall return a FAILED_PRECONDITION error.
+
+If a SetRequest call is made without extension whilst the existing rollback counter is running then a
+FAILED PRECONDITION error is returned.
+
```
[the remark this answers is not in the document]
kidiyoor: As mentioned above - `If a SetRequest call is made without extension whilst the existing rollback counter is running then a FAILED PRECONDITION error is returned.` 

We would return error as the server has no way to know if the previous configuration is good until the confirmation call is received. The proposal is NOT to prioritize SetRequest without the extension while server is waiting for confirmation.

earies:
> I do not consider "we baked this in an RFC" as a yardstick for excellent (or even implementable) technical design
> ...
> We should avoid referencing prior art as the only justification

I think there is a misunderstanding of my point.  My point is not a blind comparison to an RFC but rather drawing reference to existing standards, implementations and architectures that have been shipping for years since a majority of this work is reinventing the wheel often coming in reverse but choosing to deviate without clear reason.

IMO when reinventing the wheel, there should be an understanding of history, existing implementations and clear problem statements followed with proper justification as to the reason why a prior technology cannot suffice.

The concept in this PR is the same from prior technologies but there is a choice to change some behaviors in a less efficient fashion and that is what I am challenging here is all.  You can choose to align to other like technologies or not but let's get some good reasoning behind such.  A PR without description or prior reference is also not a good justification.

Many implementations will support _all_ prior technologies concurrently thus obvious, the more alignment, the more likelyhood of desired outcomes.  Any misalignment is new implementation which is fine but should be justified with an understanding of co-existance.

kidiyoor:
Above comments already address why this should be an extension over modifications to the core. The initial proposal did deviate from the RFC, as I was not aware of its existence. It makes sense to align with the RFC as much as possible. I have incorporated most of feedbacks and created a new proposal #197. We can probably continue discuss about the open questions in the new issue.

hellt:
Hi all,
Internally we had developed an extension for commit-confirmed that we would like to share with the wider group https://github.com/openconfig/reference/pull/198

The main differences from the proposal in #196 and #197 is the usage of `commit_id` to provide better session management and prevent concurrent access issues.

We also used a set of actions, expressed as an enum to indicate the procedure call to use.

rpc/gnmi/gnmi-commit-confirmed.md
```diff
@@ -0,0 +1,106 @@
...
+### 3.2.1 Commit
+A commit can be initiated by providing `CommitRequest` as action in the extension. A `id` must to be
+provided by the client. The server shall associate the commit with the provided `id`.
+During confirm or cancel action the provided `id` must match the `id` of the on-going commit.
+`rollback_duration` can be used to override the default rollback duration which is 10min.
+If a confirmation call is not received before the rollback duration then the configuration is reverted.
+
+If a commit is issued whilst an existing rollback counter is running then the server returns with
+FAILED_PRECONDITION error.
+
```
hellt, line 89: @dplore @kidiyoor upon further review of the spec I wanted to call out one important point that I think is missing in this proposal - **Resetting the rollback timer**.

We would like to consider adding this option to the spec of the extension. 

For network wide commits we want to set a short timeout for fast recovery but extend it while waiting for slow nodes on the network.

In our original proposal we had the following definition of the resetting logic:

> To reset the timeout for a commit for which the timeout timer has not yet expired, the client MUST include the `CommitConfirmed` extension in the `SetRequest` message without any content in `update`/`delete`/`replace` fields. The `CommitConfirmed` message MUST include the `COMMIT_CONFIRMED_ACTION_START` action, `timeout` value set to the new desired timeout, and `commit_id` field set to the commit identifier returned by the network device in the respecting `SetResponse` message.

> The network device should reset the commit rollback timeout to the new value and return the `SetResponse` message without `CommitConfirmed` extension.

kidiyoor: This looks like an new action that can be added on top of the current proposal ? Could this be added in a follow-up PR ?

hellt: While we can definitely do that in a separate PR I wanted to check why do you think this has to be done separately.

We can discuss if that warrants another message under `oneof`, or if it can be the same `CommitRequest` provided with the same ID and a new rollback timeout value.

rpc/gnmi/gnmi-commit-confirmed.md
```diff
@@ -0,0 +1,106 @@
...
+message CommitCancel {}
+```
+
+## 3.2 SetRequest handling                                                        
+
+### 3.2.1 Commit
+A commit can be initiated by providing `CommitRequest` as action in the extension. A `id` must to be
+provided by the client. The server shall associate the commit with the provided `id`.
+During confirm or cancel action the provided `id` must match the `id` of the on-going commit.
+`rollback_duration` can be used to override the default rollback duration which is 10min.
```
hellt, line 84: shall we explicitly call out that the server should maintain a default 10m interval when rollback_duration is not present?

hellt: additionally, it might be worth repeating in what units the rollback is provided. It is not easy to find the googleapis repo where the typedef is set.

rpc/gnmi/gnmi-commit-confirmed.md
hellt: I was hoping we could do this in this PR, to keep the momentum.

The resetting timeout is in-line with prior art in netconf https://datatracker.ietf.org/doc/html/rfc6241#section-8.4

> If a follow-up confirmed <commit> operation is issued before the timer expires, the timer is reset to the new value (600 seconds by default).

We can argue if that is better to be modelled as another distinctive action, or be part of the `CommitRequest` message.

I can create a PR targeting this PR if required with the proposed additions.

kidiyoor: If additional changes can be made in subsequent proposal/PR we should consider that to unblock folks and avoid unnecessary delays for already approved changes - https://github.com/openconfig/gnmi/pull/155#pullrequestreview-1731027525. 

I suspect there will be some discussion and decision to be made in terms of the syntax of the `ResetTimer` cmd. Probably best to keep the discussion in separate PR as `ResetTimer` is not core functionality of `Confirmed Commit`, its an optimization.

hellt: We can consider giving it a round of discussion at the next OC call, given how frictionless and low-engaging the past few calls went I do not expect this to bear any substantial discussions.

The reason I wanted to do this here, is that I think this rather belongs to the commit confirmed functionality, and therefore must be defined in the same spec file

rpc/gnmi/gnmi-commit-confirmed.md
kidiyoor: Updated.

rpc/gnmi/gnmi-commit-confirmed.md
dplore: The reset rollback timer is being worked at https://github.com/openconfig/gnmi/pull/161

dplore (approved):

Issue #197, opened 2023-10-09
Support confirmed commit in gNMI SetRequest


kidiyoor:
A PR https://github.com/openconfig/reference/pull/196 was created with an initial proposal to initiate discussion to add confirmed commit support. From the feedbacks received we have drafted a new proposal, creating this issue to continue discussion on the new proposal.

The proposed proto has a subset of confirmed commit functionality as defined in NETCONF protocol([RFC6241](https://datatracker.ietf.org/doc/html/rfc6241#section-8.4)). The proposal has a healthy disregard to few functionality defined in the RFC with the intention that most of the gRPC API clients are going to be automated systems and the proto should facilitate a simpler implementation of client workflows and server implementation.

The server can be viewed as a singleton resource, at any given point in time there can be only one commit active. This commit can be either confirmed or canceled before a new commit can begin. Client is expected to provide full configuration during the commit request, the commit cannot be amended once issued.

With these in mind, below are the two proposals-

### Proposal-1
Proto update PR: https://github.com/openconfig/gnmi/pull/155
Reference/README update PR: https://github.com/openconfig/reference/pull/196
```
message Commit {
  oneof action {
    CommitRequest commit = 1;
    CommitConfirm confirm = 2;
    CommitCancel cancel = 3;
  }
}
```
Above spec limits the transactions to below three transactions-
1. Allow commit
2. Allow confirm
3. Allow cancel

**Note**: It doesn’t allow a new commit along with confirmation of previous commit.

Handing of Proposal-1 is explained [here](https://docs.google.com/document/d/1pT096pavzhHzJ3k5bvJffSLVdyG3jdQmJj6JfNvOKIo/edit?usp=sharing).

Alternate discussion options:
### Proposal-2
```
message Commit {
  google.protobuf.Duration rollback_duration = 1;
  string commit_id = 2;
  string confirm_id = 3;
  string cancel_id = 4;
}
```
Handing of Proposal-2 is explained [here](https://docs.google.com/document/d/1pT096pavzhHzJ3k5bvJffSLVdyG3jdQmJj6JfNvOKIo/edit?usp=sharing).

Allows below transactions-
1. Allow commit
2. Allow confirm
3. Allow cancel
4. Allow confirm + new commit
5. Allow cancel + new commit

Open Questions
1. Is transaction 4. and 5. a valid use-case in current operations practices ?
2. Are there any anticipated use-case for 4. and 5. transactional model ?

PR #198, opened 2023-10-11
Commit confirmed gNMI extension

hellt:
Hi @robshakir, @kidiyoor 

This is a proposal for gNMI Commit Confirmed extension as a response to #196 #197 

It comes with the proposed extension proto spec in the accompanying PR https://github.com/openconfig/gnmi/pull/154

### DIFF
--- a/rpc/gnmi/gnmi-commit-confirmed.md
+++ b/rpc/gnmi/gnmi-commit-confirmed.md
@@ -1,0 +1,68 @@
+# gNMI Commit Confirmed Extension
+
+**Contributors:**
+
+**Date:** October 6, 2023
+
+**Version:** 0.1.0
+
+## 1. Purpose
+
+In certain deployments, client and server is separated by a complex network, hence we cannot assume
+
+- The pushed configuration will not break connectivity to the network device.
+- The network device has out-of-band access.
+
+This extension provides a way for a controller/operator to apply updates in a `SetRequest` that will require a subsequent confirmation in a time window specified by the `timeout` value.
+
+If no confirmation is received within the time window, the network device will revert the changes restoring the previous state. This workflow is particularly useful to ensure that erroneous configuration will not break connectivity to the network device.
+
+## 2. Definition
+
+A [`CommitConfirmed`](TBD) message is added to the list of well-known gNMI Extensions and is defined in [gnmi_ext.proto](TBD).
+
+The message contains the following fields:
+
+- `action` - The action to be taken by the network device upon receiving the `SetRequest` with the `CommitConfirmed` extension.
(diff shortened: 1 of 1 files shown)


rpc/gnmi/gnmi-commit-confirmed.md
```diff
@@ -0,0 +1,68 @@
...
+
+- `action` - The action to be taken by the network device upon receiving the `SetRequest` with the `CommitConfirmed` extension.
+- `timeout` - time in seconds/nanoseconds to wait for a confirmation before reverting the changes.
+- `commit_id` - commit identifier returned by the network device that is associated with the commit operation.
+
+### 3. Set Request handling
+
+#### 3.1 Initiating Commit Confirmed operation
+
+To initiate a Commit Confirmed operation, the client MUST include the `CommitConfirmed` extension in the `SetRequest` message. The `CommitConfirmed` message MUST include the `COMMIT_CONFIRMED_ACTION_START` action to indicate that the network device MUST apply the configuration changes and start the commit rollback timeout waiting for a confirmation action.
```
ncorran, line 34: We need to specify the locking here - i.e. the commit to commit confirm (or abort or timeout) is a window in which the server is locked from the perspective of other sessions.  I would prefer this to be explicit in the ACTION naming such as START_W_LOCK/CONFIRM_W_UNLOCK/REJECT_W_UNLOCK which may push towards:
ACTION_LOCK (instead of start), ACTION_UNLOCK_REJECT, ACTION_UNLOCK_CONFIRM.

Failing this being explicit in the actions it should be very clear in this description and in the proto etc.

rpc/gnmi/gnmi-commit-confirmed.md
```diff
@@ -0,0 +1,68 @@
...
+
+The network device should reset the commit rollback timeout to the new value and return the `SetResponse` message without `CommitConfirmed` extension.
+
+### 4. Commit ID
+
+The commit identifier returned by the network device in the `SetResponse` message with `COMMIT_CONFIRMED_ACTION_START` action enables controlled session management and prevents erroneous actions to be taken by the network device due to race conditions.
+
+For example, when commit_id is not used it is theoretically possible that the commit confirmed action will be processed by the network device after the timeout has expired, causing erroneous action to be taken. Or multiple actors will attempt to perform different actions on the same commit.
```
ncorran, line 68: Add some comment about security here - is the commit-id sufficient to really identify the initial commit owner and are we happy with that? (vs constraining this to within a session so we can check some session params on the server as well as the id? or e.g. store and check the AAA credentials of the user vs the initiator??)

hellt:
replaced by #196

PR #203, opened 2024-01-26
added set rollback duration documentation

hellt:
Accompanying PR for https://github.com/openconfig/gnmi/pull/164

/cc @kidiyoor

### DIFF
--- a/rpc/gnmi/gnmi-commit-confirmed.md
+++ b/rpc/gnmi/gnmi-commit-confirmed.md
@@ -130,6 +130,10 @@ When a commit is on-going, the existing rollback duration can be reset to a new 
 by providing `SetRollbackDurationRequest` as action in the extension. The value of `id` should
 be equivalent to the id of the on-going commit.
 
+Note, that the rollback duration value provided in the `SetRollbackDurationRequest` action
+will effectively overwrite the existing rollback duration timer. It will not append to the
+existing rollback duration time but set it to the new value provided.
+
 The value of `rollback_duration` should be provided and be greater than 0. If the value is 0 or not
 provided then INVALID_ARGUMENT error is returned.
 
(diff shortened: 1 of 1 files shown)


rpc/gnmi/gnmi-commit-confirmed.md
```diff
@@ -103,8 +117,20 @@ If the server is not waiting for confirmation or if the value doesn’t match th
...
+
 Cancellation can be issued by providing `CancelRequest` as action in the extension. The value of `id`
 should be equivalent to the id of the on-going commit.
 
 If the server is not waiting for cancellation or if the value doesn’t match the on-going commit
 then FAILED_PRECONDITION or INVALID_ARGUMENT error is returned respectively.
+
+### 3.2.4 Set Rollback Duration
+
+When a commit is on-going, a new rollback duration can be set by providing `SetRollbackDurationRequest`
```
kidiyoor, line 129: Could you make it clear if the existing rollback counter will be reset or if the new duration value value will be added on to the on-going counter ?

kidiyoor:
LGTM

rpc/gnmi/gnmi-commit-confirmed.md
hellt: done in 19cd325

LimeHat: Perhaps a similar comment, but I think we need a bit more clarity here.

Can you please explicitly specify the "starting point" of the new timer (do we count the duration from the initial commit timestamp, or from 0 after receiving `SetRollbackDurationRequest`)?

hellt: done in [0c1d8f4](https://github.com/openconfig/reference/pull/203/commits/0c1d8f4216e0f51b4994f0053305e7933696c35f)

kidiyoor (approved):

hellt:
@dplore since the proto for this feature has been merged, can we merge also the doc part?

dplore (approved):

PR #215, opened 2024-08-29
Clarify dynamic, bootz, boot_config and config order

dplore:


rpc/gnmi/gnmi-union_replace.md
```diff
@@ -494,16 +494,23 @@ union_replace: {
...
-gNSI configuration.
+order is dynamic configuration, followed by [boot_config](https://github.com/openconfig/bootz/blob/e0eb604d8e7e089ecf0d4553cce79de769328173/server/entitymanager/proto/entity.proto#L55)
+configuration, followed by gNSI configuration.  That is, first apply dynamic
+configuration, then apply the boot_config (set by bootz) which may override
+dynamic configuration. Finally, apply gNSI configuration which may override
+dynamic and boot_config.
+
+Note that bootz may [configure gNSI values](https://github.com/openconfig/bootz/blob/e0eb604d8e7e089ecf0d4553cce79de769328173/server/entitymanager/proto/entity.proto#L32).
+This is treated as gNSI config, using the gNSI interface.  The order of
+config application still follows the dynamic -> boot_config -> gnsi order.
```
marcushines, line 513: The only maybe other question would be how gnoi.BootConfig.SetBootConfig interacts but it is actually pretty simple it works just like bootz would have - it overwrites everything that is present in that message (generally) ONLY boot_config would be present in the message as only super rarely would you need to update anything else (as the normal gnsi endpoints should used for those operations)

rpc/gnmi/gnmi-union_replace.md
```diff
@@ -494,16 +494,24 @@ union_replace: {
...
 expected to be modified while the NOS is running.   Security configuration is a
 special case of dynamic configuration and is managed by gNSI.  Configuration
 necessary to boot and make a factory default NOS manageable by gNMI is defined
 as bootz configuration.
 
-bootz configuration items can be expressed as OC, NY and CLI. Bootz defines an
-order (or precedence) to apply the configuration name spaces to the NOS.  The
-order is dynamic configuration, followed by bootz configuration, followed by
-gNSI configuration.
+bootz [boot_config](https://github.com/openconfig/bootz/blob/e0eb604d8e7e089ecf0d4553cce79de769328173/server/entitymanager/proto/entity.proto#L55)
```
LimeHat, line 503: the references should point to the bootz.proto https://github.com/openconfig/bootz/blob/e0eb604d8e7e089ecf0d4553cce79de769328173/proto/bootz.proto

rpc/gnmi/gnmi-union_replace.md
dplore: Added this point as an example scenario.

rpc/gnmi/gnmi-union_replace.md
dplore: Nice catch, fixed.

PR #223, opened 2025-04-07
Add clarification as to error handling with gNMI extensions.

robshakir:
```
 * (M) rpc/gnmi/gnmi-extensions.md
  - Add clarification that unknown extensions must not be treated
    as an error.
  - Add clarification that known extensions that are invalid may
    be treated as an error.
```

### DIFF
--- a/rpc/gnmi/gnmi-extensions.md
+++ b/rpc/gnmi/gnmi-extensions.md
@@ -3,9 +3,9 @@
 **Contributors**: Rob Shakir (robjs@google.com), Carl Lebsack (csl@google.com),
 Nick Ethier (nethier@jive.com), Anees Shaikh (aashaikh@google.com)
 
-**Updated**: January 25th, 2018
+**Updated**: April 7, 2025
 
-**Version**: 0.1.0
+**Version**: 0.2.0
 
 ## Extending gNMI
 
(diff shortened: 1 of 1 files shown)


robshakir:
@karimra @limehat @dplore @ncorran @earies @aaronbee PTAL

LimeHat:
Lgtm. Thanks.

earies (approved):
LGTM

rpc/gnmi/gnmi-extensions.md
dplore, line 1: Content LGTM.  

Also update the revision date and version?

Should we also update to the [gnmi spec](https://github.com/openconfig/reference/blob/9f5820d75b7840f257669c1c5e7dde56d607c135/rpc/gnmi/gnmi-specification.md?plain=1#L610)  and [gnmi_ext.proto](https://github.com/openconfig/gnmi/blob/master/proto/gnmi_ext/gnmi_ext.proto)?

robshakir: Updated revision date + version.

What do you propose we update in the gNMI spec + proto? Part of the reasons for the modularity of specs we have is to prevent there being a single doc that is somewhat unmaintainable.

robshakir: @dplore friendly ping.

PR #227, opened 2025-07-16
Update gnmi-specification.md Target_Defined mode description

ashu-ciena:
Add clarity on using sample-interval in target defined mode.

### DIFF
--- a/rpc/gnmi/gnmi-specification.md
+++ b/rpc/gnmi/gnmi-specification.md
@@ -1638,12 +1638,10 @@ with one of the following `modes`:
     driven (e.g., the changing of state of an entity based on an external
     trigger) then an `ON_CHANGE` subscription may be created, whereas if other
     data represents counter values, a `SAMPLE` subscription may be created.
-    When the target chooses a `SAMPLE` type of subscription for a leaf and the
-    client does not specify the `sample_interval` then target must choose a
-    `sample_interval` too. In this scenario, The target MAY change the
-    `sample_interval` dynamically without notifying the client. If the client does
-    include a `sample_interval` in the subscription request, then the same rules
-    for processing a `SAMPLE` mode of subscription applies.
+    The target is expected to choose and change a `sample_interval` dynamically.
+    If `sample_interval` is specified in the subscribe request, the target
+    `SHOULD` reject the subscription by closing the Subscribe RPC specifying an
+    `InvalidArgument (3)` error code.
 
 ##### 3.5.1.5.3 POLL Subscriptions
 
--- a/rpc/gnmi/gnmi-specification.md
+++ b/rpc/gnmi/gnmi-specification.md
@@ -1638,9 +1638,12 @@ with one of the following `modes`:
     driven (e.g., the changing of state of an entity based on an external
     trigger) then an `ON_CHANGE` subscription may be created, whereas if other
     data represents counter values, a `SAMPLE` subscription may be created.
-    In addition to this, the target MAY decide the `sample_interval` for a
-    `SAMPLE` type of subscription under this mode, on a per-leaf basis. The target
-    MAY change the `sample_interval` dynamically without notifying the client.
+    When the target chooses a `SAMPLE` type of subscription for a leaf and the
+    client does not specify the `sample_interval` then target must choose a
+    `sample_interval` too. In this scenario, The target MAY change the
+    `sample_interval` dynamically without notifying the client. If the client does
+    include a `sample_interval` in the subscription request, then the same rules
+    for processing a `SAMPLE` mode of subscription applies.
 
 ##### 3.5.1.5.3 POLL Subscriptions
 
--- a/rpc/gnmi/gnmi-specification.md
+++ b/rpc/gnmi/gnmi-specification.md
@@ -978,7 +978,7 @@ The `GetResponse` message consists of:
 
 The `Get` RPC is intended for clients to retrieve relatively small sets of data
 as complete objects, for example a part of the configuration. Such requests are
-not expected to put a signifit resource burden on the target. Since the
+not expected to put a significant resource burden on the target. Since the
 target is expected to return the entire snapshot in the `GetResponse` message,
 `Get` is not well-suited for retrieving very large data sets, such as the full
 contents of the routing table, or the entire component inventory. For such
@@ -1022,7 +1022,7 @@ A SetRequest containing
 [union_replace](https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-union_replace.md)
 operations MUST NOT contain delete, replace and update operations.
 
-The order of the fields MUST be treated as signifit within a
+The order of the fields MUST be treated as significant within a
 single `SetRequest` message.  If a single path is specified multiple times for a
 single operation (i.e., within `update` or `replace`), then the state of the
 target MUST reflect the application of all of the operations in order, even if
@@ -1034,7 +1034,7 @@ In response to a `SetRequest`, the target MUST respond with a `SetResponse`
 message. For each operation specified in the `SetRequest` message, an
 `UpdateResult` message MUST be included in the response field of the
 `SetResponse`. The order in which the operations are applied MUST be maintained
-such that `UpdateResult` messages  be correlated to the `SetRequest`
+such that `UpdateResult` messages can be correlated to the `SetRequest`
 operations. In the case of a failure of an operation, the status of the
 `UpdateResult` message MUST be populated with error information as per the
 specification in [Section 3.4.7](#347-error-handling). In addition, the status
@@ -1327,14 +1327,14 @@ operation within the `SetRequest` the target MUST set the `message` field of the
 For the operation that the target is unable to process, the code indicated in
 the status message returned within the RPC response MUST be set to a specific
 error code indicating the reason for failure based on the following mappings to
-onical gRPC error codes:
+canonical gRPC error codes:
 
 - When the client has specified metadata requiring authentication (see [Section
     3.1](#31-session-security-authentication-and-rpc-authorization)), and the
     authentication fails -  `Unauthenticated (16)`.
 - When the client does not have permission to modify the path specified by the
     operation - `PermissionDenied (7)`.
-- When the operation specifies a path that not be parsed by the target -
+- When the operation specifies a path that cannot be parsed by the target -
     `InvalidArgument (3)`. In this case, the `message` field of the returned
     status specified SHOULD specify human-readable text indicating that the path
     could not be parsed.
@@ -1370,8 +1370,8 @@ The target generates messages according to the type of subscription that has
 been created, at the frequency requested by the client. The methods to create
 subscriptions are described in [Section 3.5.1](#351-managing-subscriptions).
 
-Subscriptions are created for a set of paths - which not be modified
-throughout the lifetime of the subscription. In order to cel a subscription,
+Subscriptions are created for a set of paths - which cannot be modified
+throughout the lifetime of the subscription. In order to cancel a subscription,
 the client cancels the `Subscribe` RPC associated with the subscription, or
 terminates the entire gRPC session.
 
--- a/rpc/gnmi/gnmi-specification.md
+++ b/rpc/gnmi/gnmi-specification.md
@@ -978,7 +978,7 @@ The `GetResponse` message consists of:
 
 The `Get` RPC is intended for clients to retrieve relatively small sets of data
 as complete objects, for example a part of the configuration. Such requests are
-not expected to put a significant resource burden on the target. Since the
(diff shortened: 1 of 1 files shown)


ashu-ciena (commented):
The changes appearing in this PR, in other than target defined mode section, are not done by me. In my copy of the fork, these changes are already fixed. So, I am not sure why these unrelated changes are appearing in this PR.

ashu-ciena:
@dplore Can you please add relevant reviewers to this PR? Any information missing in the PR ?

dplore:
I've added this to the queue for OC Operator review and pinged a couple of individual reviewers.

Making this mandatory would make it a breaking change to the base GNMI specification which we should not allow.   

I don't think we expect or require an sample_interval for TARGET_DEFINED.   The target may choose it's own sample_interval.   I can see this as possibly an optional field for TARGET_DEFINED subscriptions.

rpc/gnmi/gnmi-specification.md
```diff
@@ -1638,6 +1638,10 @@ with one of the following `modes`:
     driven (e.g., the changing of state of an entity based on an external
     trigger) then an `ON_CHANGE` subscription may be created, whereas if other
     data represents counter values, a `SAMPLE` subscription may be created.
+    In addition to this, the target MUST select the `sample_interval` for a
```
robshakir: Agreeing with some of Darren's comments, and adding my own:

* Making this required is backwards incompatible, we should avoid breaking existing clients.
* `TARGET_DEFINED` is explicitly a loose(r) contract with the device -- as a client you are letting the device choose how best to send data to you based on its implementation status. Setting the `sample_interval` seems to have two big problems to me:
  * we are expecting the client to understand what `sample_interval` is possible for the device -- whereas the device itself is meant to be making choices about what is possible for _it_ within this mode.
  * if we set a `sample_interval` for a subscription to `/`, say, and the target has different supportable rates per subtree (e.g., data plane counters can be collected every 10 seconds, control plane can be collected every 1 second) -- then what do we set this value to? If we set it to 1, then is the subscription rejected? If not, what does this included `sample_interval` actually specify?

Generally, I feel that this change is trying to make a contract that is loose more explicit -- at which point a more targeted subscription (i.e., NOT using `TARGET_DEFINED`) would probably be more suitable in order to set a clear target<->client expectation that is testable and can be relied upon in production.

rpc/gnmi/gnmi-specification.md
```diff
@@ -1638,6 +1638,10 @@ with one of the following `modes`:
...
+    `SAMPLE` subscription under this mode, on a per-leaf basis for a fast
+    periodic response. The target MAY change the `sample_interval` dynamically
+    without notifying the client.
```
robshakir: Are you proposing any constraints on this? Can the target back off the `sample_interval` to one sample per hour, or per day? It seems like there is significant chance of breaking the intended behaviour of the wider network system without some constraints here.

Practically, how would this mechanism be triggered? How does a target decide that backing off the sample interval is the right thing to do and by how much?

rpc/gnmi/gnmi-specification.md
ashu-ciena: Thanks for the comments. I agree to remove the sample interval clause from the update and let the contract loose to be backward compatible and to keep the spirit of the target-defined mode.

dplore:
Reviewed in July 29, 2025 OC Operators meeting.   

An additional idea is that this "if the target decides to use sample, the client suggests a sample interval" idea should be implemented in a new gnmi subscription mode.  This way the target can communicate if the "suggested sample interval" feature is supported or not.

I think this PR would benefit from a more information about the operational use cases we are trying to solve for before we can provide better feedback of what should be implemented.

rpc/gnmi/gnmi-specification.md
ashu-ciena: Removed the MUST clause in latest upload. Can you please review again.

ashu-ciena:
> Reviewed in July 29, 2025 OC Operators meeting.
> 
> An additional idea is that this "if the target decides to use sample, the client suggests a sample interval" idea should be implemented in a new gnmi subscription mode. This way the target can communicate if the "suggested sample interval" feature is supported or not.
> 
> I think this PR would benefit from a more information about the operational use cases we are trying to solve for before we can provide better feedback of what should be implemented.

Are you suggesting a new subscription mode? I was wondering if the user subscribes in target-defined mode and also provides the sample interval, the target can then communicate if the "suggested sample interval" is viable or not on the subscribed xpath. If , there is no sample interval provided, target can choose.

ashu-ciena:
> > Reviewed in July 29, 2025 OC Operators meeting.
> > An additional idea is that this "if the target decides to use sample, the client suggests a sample interval" idea should be implemented in a new gnmi subscription mode. This way the target can communicate if the "suggested sample interval" feature is supported or not.
> > I think this PR would benefit from a more information about the operational use cases we are trying to solve for before we can provide better feedback of what should be implemented.
> 
> Are you suggesting a new subscription mode? I was wondering if the user subscribes in target-defined mode and also provides the sample interval, the target can then communicate if the "suggested sample interval" is viable or not on the subscribed xpath. If , there is no sample interval provided, target can choose.

@dplore and @robshakir do you want more discussion on this in Thursday meeting? If not, can we give our final set of comments on the latest upload and we can proceed to merge?

ashu-ciena:
Following up on this thread, are there any open questions/comments? This PR requires approval, looking forward to it.

rpc/gnmi/gnmi-specification.md
```diff
@@ -1638,6 +1638,9 @@ with one of the following `modes`:
     driven (e.g., the changing of state of an entity based on an external
     trigger) then an `ON_CHANGE` subscription may be created, whereas if other
     data represents counter values, a `SAMPLE` subscription may be created.
+    In addition to this, the target MAY decide the `sample_interval` for a
+    `SAMPLE` type of subscription under this mode, on a per-leaf basis. The target
+    MAY change the `sample_interval` dynamically without notifying the client.
```
dplore: Here's a suggestion to slightly modify the wording regarding sample_interval.

When the target chooses a `SAMPLE` mode subscription for a leaf, if the client did not specify 
the `sample_interval` then target must choose a `sample_interval`.  In this scenario, The target
MAY change the `sample_interval` dynamically without notifying the client.  If the client does 
include a `sample_interval`, then the same rules for processing a `SAMPLE` mode
subscription apply.

hellt: I am also cautious of the original substring that went like

> In addition to this, the target MAY decide the `sample_interval` for a
`SAMPLE` type of subscription under this mode, **on a per-leaf basis.**

I think there is no such thing as per-leaf basis, as all leafs are scoped to a subscription list.

In your suggestion @dplore I don't see this, so I presume you left it out on purpose as well

ashu-ciena: @dplore, so is your suggestion to update both sample and target-defined sections with the additional info on sample interval equals to 0?

ashu-ciena: And as I understand the sample mode of subscription, it mandates specifying the sample interval. If client doesn't, the request should be rejected with invalid argument error.

ashu-ciena:
I am note sure why this PR is in "Waiting For Author" state? I think it should be moved to "Ready to discuss" or "In Progress" mode.  I have incorporated all the comments, I received so far.

ElodinLaarz:
Moving to ready-to-discuss.

The ability to specify an interval when setting the value of `TARGET_DEFINED` is surprising to me? But we'll see what folks at the Operators Meeting think.

dplore:
Reviewed in Oct 14, 2025 OC Operators meeting.

General feedback is:
- When using `TARGET_DEFINED`, a target is  expected to choose and change a sample interval dynamically, although this is undefined in the specification.  We would be happy to accept an update to the specification that this is the expectation.

- There's no expectation that a client can set a `sample_interval` in a `TARGET_DEFINED` subscription mode.  The gnmi.proto specifies that `sample_interval` is used in `SAMPLE` mode ([ref](https://github.com/openconfig/gnmi/blob/891c5a8cdcd670528157283cde220adb0f43f63e/proto/gnmi/gnmi.proto#L303)).

earies:
>     * There's no expectation that a client can set a `sample_interval` in a `TARGET_DEFINED` subscription mode.  The gnmi.proto specifies that `sample_interval` is used in `SAMPLE` mode ([ref](https://github.com/openconfig/gnmi/blob/891c5a8cdcd670528157283cde220adb0f43f63e/proto/gnmi/gnmi.proto#L303)).

Suggestion for discussion: If `sample_interval` is set to 0, have the element choose acceptable intervals.  If set to non zero, treat the same as sample mode where the device can choose to honor/error

ashu-ciena:
@dplore and @earies the text in the current state of spec is only focusing on the type of subscription in target defined mode. So, we have the liberty to add clarifications regarding the sample interval, if target chose to treat the subscription as sample mode. Not specifying the sample interval is equivalent to specifying it as "0" value.  The Subscription proto is same for both sample and target defined mode, so we can not restrict users to provide a sample interval even in target defined mode.  I somewhat agree that if user does, target can choose to accept or reject the subscription with "Invalid Parameter" error. 

If you all agree, I can update my PR with these clarifications.

dplore:
The issue is the currently proposed text to be added for TARGET_DEFINED conflicts with the gnmi.proto field definition for `sample_interval`.    (which is: `uint64 sample_interval = 3;  // ns between samples in SAMPLE mode.`).    

It is ok to propose text that says use of `sample_interval` is only defined for `Subscription.mode` is `SAMPLE`.    If `sample_interval` is used in Subscription.mode TARGET_DEFINED , the target should reject the subscription with grpc status code `INVALID_ARGUMENT`.

dplore:
If you are very interested in allowing a client to optionally choose a `sample_interval` when using TARGET_DEFINED mode  a gnmi extension might be a better place for this.  We have managed to not make any changes to the core gnmi protocol in some time and would like to have any additions appear as extensions.

raghubk:
> If you are very interested in allowing a client to optionally choose a `sample_interval` when using TARGET_DEFINED mode a gnmi extension might be a better place for this. We have managed to not make any changes to the core gnmi protocol in some time and would like to have any additions appear as extensions.

With TARGET_DEFINED, as the name suggests the "TARGET" should define the sample interval as well. Implementors may choose to use "default" sample interval in TARGET_DEFINED mode, just like when sample_interval is set to zero in SAMPLE mode. Thus no change to the gnmi notification message is required either.

ashu-ciena:
I have updated the PR with the suggested text. Please do a final review.
@raghubk if sample interval is not provided, it is equivalent to 0, so just like sample mode, target can choose the sample interval. In contrast to sample mode of subscription with sample interval 0, the target is not bound to choose the lowest minimum possible value of sample interval imo.

ashu-ciena:
@dplore , resolved the conflict and the PR is updated with the latest outcome of Oct 14, 2025 OC Operators meeting and inline with proto definition that sample interval is for SAMPLE mode of subscription. Can you move this to In-progress state?

dplore:
Will review in Mar 24, 2026 OC Operators meeting

rpc/gnmi/gnmi-specification.md
```diff
@@ -1636,6 +1636,10 @@ with one of the following `modes`:
   driven (e.g., the changing of state of an entity based on an external
   trigger) then an `ON_CHANGE` subscription may be created, whereas if other
   data represents counter values, a `SAMPLE` subscription may be created.
+  The target is expected to choose and change a `sample_interval` dynamically.
```
LimeHat: I disagree with the proposed wording. There should be no such expectation.
I suggest using "may change ... dynamically` instead.

dplore: Good comment, this update should specify "MAY", "SHOULD" or "MUST" terms for this.   I agree with the "MAY" language.

ashu-ciena: updated the PR with the proposed change.

ashu-ciena:
PR is updated with outcome of Mar 24, 2026 OC Operators meeting and it is in its final stage now for merge. Need mandatory approvals to proceed.

dplore:
Will review in Apr 14, 2026 OC Operators meeting

ashu-ciena:
> Will review in Apr 14, 2026 OC Operators meeting

Is there any meeting invite so that I can join?

ElodinLaarz:
Discussed at OC Operators Meeting April 14th, 2026:

The language update makes sense and matches the understanding of how TARGET_DEFINED is defined. Can move this into last-call in 2 weeks and see if there are any objections.

ElodinLaarz (approved):

dplore (approved):

PR #229, opened 2025-10-01
gNMI: Add section to cover atomic datasets

earies:
Addition of subsection to `3.5.2.1 Bundling of Telemetry Updates` that covers
atomic datasets and when there is a relationship to YANG modeled data.

Reference: https://github.com/openconfig/gnmi/issues/180

### DIFF
--- a/rpc/gnmi/gnmi-specification.md
+++ b/rpc/gnmi/gnmi-specification.md
@@ -4,10 +4,10 @@
 Paul Borman, Marcus Hines, Carl Lebsack, Chris Morrow, Anees Shaikh, Rob Shakir, Wen Bo Li, Darren Loher
 
 **Date:**
-May 25, 2023
+Oct 1, 2025
 
 **Version:**
-0.10.0
+0.11.0
 
 **[gNMI service](https://github.com/openconfig/gnmi/blob/master/proto/gnmi/gnmi.proto) compatibility:**
 0.10.x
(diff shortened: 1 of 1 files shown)


rpc/gnmi/gnmi-specification.md
```diff
@@ -1680,6 +1680,44 @@ of the values, retaining a consistent sample period and having a robust
...
+
+When multiple datapoints (e.g. Update messages) are bundled within a single
+Notification message and the intent is that the dataset must be delivered as a
+single PDU (e.g. in order to make sense of relational data), the `atomic` field
+in the Notification message structure MUST be set to `true` and all datapoints
+will share the same Notification timestamp.
+
+If the data content is modeled in the YANG language, a YANG extension SHOULD be
+annotated on a corresponding structure to be delivered atomically.  Valid YANG
+structures are: `[ list, container ]` and scale MUST be taken into
```
rgwilton, line 1693: We need go consider atomic list entries vs the whole list is atomic.  I suspect that this would need different annotations.

earies: Good point to call out.  So far, I believe every use-case of atomicity to date has been on a list entry level (each entry can be treated as an atomic unit) - I am currently unaware of a scenario in which _all_ list entries would be mandated to be delivered atomically but it is very well possible.

To the point of another possible annotation, what if we cleaned up the verbiage here to something like:

```
If the data content is modeled in the YANG language, a YANG extension SHOULD be
annotated on a corresponding structure to be delivered atomically.  Valid YANG
structures are: `[ list, container ]` and scale MUST be taken into consideration with
respect to atomic annotations.

When an annotation exists on a container, all child nodes of that container are to be
delivered atomically within a single `Notification` message.

When an annotation exists on a list, all child nodes of a single list entry are to be
delivered atomically within a single `Notification` message.  Delivering _all_ list entries
atomically is not currently supported.  Should the need arise, a future extension to
annotate at the list level can be taken into consideration.
```

Then follow-up w/ a description update to the `oc-ext:telemetry-atomic` extension detailing out this same ruleset w/ a reference to the spec section.

If we encounter a use-case of wanting to attach an annotation to a list to indicate _all_ entries to be delivered atomically, this would then call for a new extension + spec update in the future.

WDYT?

romeyod:
Reference to documentation update: https://github.com/openconfig/public/pull/1238/files

PR #232, opened 2025-10-13
update gnmi documentation on website

wisotzky:
target is to fix:
https://github.com/openconfig/gnmi/issues/227

### DIFF
--- a/ietf/drafts/draft-openconfig-rtgwg-gnmi-spec-01.xml
+++ b/ietf/drafts/draft-openconfig-rtgwg-gnmi-spec-01.xml
@@ -269,11 +269,86 @@
           <date month="January" year="2018" />
         </front>
       </reference>
+      <reference anchor="GNMI-CONFIG-SUBS" target="https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-config-subscriptions.md">
+        <front>
+          <title>gNMI Config Subscription Extension</title>
+          <author><organization>OpenConfig operator working group</organization></author>
+          <date month="July" year="2024" />
+        </front>
+      </reference>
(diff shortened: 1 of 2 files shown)


ietf/drafts/draft-openconfig-rtgwg-gnmi-spec-01.xml
```diff
@@ -269,13 +269,88 @@
           <date month="January" year="2018" />
         </front>
       </reference>
+      <reference anchor="GNMI-CONFIG-SUBS" target="https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-config-subscriptions.md">
```
earies, line 272: This .xml source for the long expired draft should not be touched unless there is intention to bring this back through the IETF (which I believe is highly unlikely nor of interest at this point) - essentially these old references can stay untouched for historical reasons is all imo.

wisotzky: Hej Ebben (@earies),

My main concern is that the documentation hosted on openconfig.net feels quite incomplete. While the Markdown files for gNMI extensions exist in the GitHub repository, they don’t seem to be used when rendering the website.

My assumption was that the XML document serves as an index of relevant files—similar to how mkdocs.yml works—and that it’s used in the CI flow to publish the site. But perhaps I missed something.

I’d be happy to help improve the publishing setup and ensure the docs are properly surfaced on the webpage. Pointing people directly to subfolders in the GitHub repo doesn’t feel like the right long-term solution.

/wiso

earies: Agreed - outside of the specification .md, I think doc reordering, a proper index and cleanup is necessary.... YMMV with the other docs atm.

This .xml is just the source to generate a draft for submission through the IETF so it's historical and not for the intentions you suggest.  Worth keeping around but not to be edited.

PR #237, opened 2026-03-05
add atomic flag parsing semantics

### LINKED ISSUE #1376: YANG Atomic extensions and behavioral expectations
While there is still unmerged work and discussion surrounding YANG atomic extensions here: https://github.com/openconfig/public/pull/1330

As well as prior discussion on the relationship between the YANG atomic extension and gNMI atomic field here: https://github.com/openconfig/gnmi/issues/180

If we use the OC AFT modeling as an example of atomic datasets

* There is 0 mention of atomic behavior expectations from a gNMI specification perspective
* These datasets are not marked as `oc-ext:telemetry-on-change` (Similar to what is described in https://github.com/openconfig/public/issues/1375)
* There is no description of expectations should one choose to query/subscribe to a subset of data under an atomic marked container/list

The intention of this issue is to spark discussion to come to consensus on expectations when

* Subscriptions target subsets of data under atomic flagged content (deliver subset, error, or entire atomic flagged content)
* YANG containers/lists are marked as atomic, agreement that this corresponds to a case when gNMI marks `atomic = true`

ElodinLaarz:
Explain how the atomic flag works and interacts with oc-ext:atomic extension.

Copilot Summary:

This pull request updates the `gnmi-specification.md` to clarify and formalize the parsing and behavioral semantics of atomic notifications in the gNMI protocol. The most significant changes include a new section detailing how to parse atomic notifications, guidance for both clients and devices on handling atomic containers in subscriptions, and an update to the revision history to document these changes.

**Atomic Notification Parsing and Behavior:**

* Added a new section, "Parsing Atomic gNMI Notifications," which details how the `atomic` flag in a `Notification` message should be interpreted, emphasizing schema-independence, implicit deletion of prior state, invalidation rules, and the importance of prefix scope.
* Clarified that the `atomic` field in the `Notification` message indicates an atomic update, and explained how atomic and non-atomic notifications affect the validity of cached state for a given prefix.

**Atomic Containers in Subscriptions and Requests:**

* Added a section on "Atomic Containers in Subscribe and Get Requests," explaining how devices should generate atomic notifications when clients subscribe to only a subset of leaves within an atomic container, and how clients should interpret such notifications.

**Documentation and Navigation Updates:**

* Updated the table of contents to include new sections for parsing atomic notifications and atomic containers in subscriptions/requests. [[1]](diffhunk://#diff-dbc0e90bf50edde03d6f90fb9f8ca59a129de12f7b77de5b3ae8912218edf915R22) [[2]](diffhunk://#diff-dbc0e90bf50edde03d6f90fb9f8ca59a129de12f7b77de5b3ae8912218edf915R77)
* Updated the revision history to version 0.11.0, documenting the new definitions for parsing semantics of the `atomic` flag.

### DIFF
--- a/rpc/gnmi/gnmi-specification.md
+++ b/rpc/gnmi/gnmi-specification.md
@@ -19,6 +19,7 @@ May 25, 2023
 - [1 Introduction](#1-introduction)
 - [2 Common Message Types and Encodings](#2-common-message-types-and-encodings)
   - [2.1 Reusable Notification Message Format](#21-reusable-notification-message-format)
+    - [2.1.1 Parsing Atomic gNMI Notifications](#211-parsing-atomic-gnmi-notifications)
   - [2.2 Common Data Types](#22-common-data-types)
     - [2.2.1 Timestamps](#221-timestamps)
     - [2.2.2 Paths](#222-paths)
@@ -73,6 +74,7 @@ May 25, 2023
       - [3.5.2.1 Bundling of Telemetry Updates](#3521-bundling-of-telemetry-updates)
       - [3.5.2.3 Sending Telemetry Updates](#3523-sending-telemetry-updates)
       - [3.5.2.4 SubscribeResponse Behavior Table](#3524-subscriberesponse-behavior-table)
+      - [3.5.2.5 Atomic Containers in Subscribe and Get Requests](#3525-atomic-containers-in-subscribe-and-get-requests)
 - [4 Appendix: Current Protobuf Message and Service Specification](#4-appendix-current-protobuf-message-and-service-specification)
 - [5 Appendix: Current Outstanding Issues/Future Features](#5-appendix-current-outstanding-issuesfuture-features)
 - [6 Copyright](#6-copyright)
@@ -142,7 +144,7 @@ The fields of the Notification message are as follows:
   definition in [2.2.1](#221-timestamps).
 - `prefix` - a prefix which is applied to all path fields (encoded as per
   [2.2.2](#222-paths)) included in the `Notification` message. The paths
-  expressed within the message are formed by the concatenation of `prefix + path`. The `prefix` always precedes the `path` elements. Further semantics
+  expressed within the message are formed by the concatenation of `prefix +   path`. The `prefix` always precedes the `path` elements. Further semantics
   of prefixes are described in [2.4.1](#241-path-prefixes).
 - `update` - a list of update messages that indicate changes in the underlying
   data of the target. Both modification and creation of data is expressed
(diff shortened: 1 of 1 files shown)


rpc/gnmi/gnmi-specification.md
```diff
@@ -165,10 +167,114 @@ The fields of the Notification message are as follows:
...
+```
+
+**2. Prior Non-Atomic Notifications**
+
+```yaml
+Prefix: a/b
+Update: some_counter
+time: 121  # Disregarded: Overridden by the atomic notification at time 123
+```
+
```
romeyod, line 245: Can we also add some documentation about only delete notifications dont need atomic flag set as such notifications with Prefix are atomic (i.e. Delete the entire subtree)

romeyod (commented):
should we add some guidance about how atomic containers should ideally not have frequently changing leafs/containers like counters?

rpc/gnmi/gnmi-specification.md
```diff
@@ -165,10 +167,114 @@ The fields of the Notification message are as follows:
...
+```yaml
+Prefix: a/b
+Update: c/d
+Update: c/e
+Update: f/g
+atomic: True
+time: 123
+```
+
```
dplore, line 196: Add a more simple example, ie:

prefix a/b
update: c/d
update: c/e
time: 1


prefix a/b
update: c/e
time: 2

Result is "a/b/c/d" is deleted.

rpc/gnmi/gnmi-specification.md
```diff
@@ -165,10 +167,114 @@ The fields of the Notification message are as follows:
...
+```yaml
+Prefix: a/b
+Update: f/g
+atomic: False # (or omitted)
+time: 124 
+```
+
+When the notification at `time: 124` is received, the atomic snapshot established at `time: 123` is invalidated. The parser must treat the tree under `a/b` as an evolving, non-atomic dataset from this point forward, until a new `atomic: True` notification is received for that prefix.
+
+In practice, it is not recommended to mix atomic and non-atomic notifications for the same prefix, as it can lead to unexpected behavior, as noted above.
```
dplore, line 275: Let's say something like:  
For OpenConfig data models, this behavior is not allowed.   If a path is marked atomic in OpenConfig data models, the server must always send atomic updates.   However, since gnmi is schema agnostic, this behavior might be useful for other schemas.

rpc/gnmi/gnmi-specification.md
```diff
@@ -165,10 +167,114 @@ The fields of the Notification message are as follows:
...
+
+```yaml
+Prefix: a
+Update: b/c/d
+Update: b/c/e
+Update: b/f/g
+atomic: True
+time: 123
+```
+
```
dplore, line 301: Seems like we might need an example added after this one where a subcontainer is updated with atomic true.   ie:

```
Prefix: a/b
Update: c/d
atomic: True
time: 124
```

Expected client behavior is this would invalidate the notification at time:123

rpc/gnmi/gnmi-specification.md
```diff
@@ -1746,6 +1852,28 @@ The following table clarifies the target behaviors for `Subscribe` for certain s
...
+If the client subscribes strictly to `a/b/c/d`, the device should generate the following notification:
+
+```yaml
+Prefix: a/b
+Update: c/d
+# c/e and f/g are omitted because they are outside the subscription scope
+atomic: True
+time: 126
+```
+
```
dplore, line 1916: Add option that server may return a gRPC error if a subscription to a sub container is not supported.

rpc/gnmi/gnmi-specification.md
ElodinLaarz: Added a section on deletes and atomic notifications.

rpc/gnmi/gnmi-specification.md
ElodinLaarz: Changed language to SHOULD and said that servers MAY instead return an error.

ElodinLaarz:
> should we add some guidance about how atomic containers should ideally not have frequently changing leafs/containers like counters?

Added a comment on best practices and frequently changing leaves.

dplore (approved):
Reviewed in March 17, 2026 OC Operators meeting without objection.  Setting last call to March 31, 2026

PR #238, opened 2026-03-05
Update JSON_ETF encoding rules in gnmi specification

dplore:
Added emphasis that JSON_IETF must use namespace qualified names where required as we observe a variety of implementations do not strictly adhere to section 4 of RFC7951.

### DIFF
--- a/rpc/gnmi/gnmi-specification.md
+++ b/rpc/gnmi/gnmi-specification.md
@@ -479,7 +479,9 @@ used. This results in quoted string values, and unquoted number values.
 described in [RFC7951](https://tools.ietf.org/html/rfc7951). Data specified with
 a type of JSON MUST be valid JSON, but no additional constraints are placed upon
 it. An implementation MUST NOT serialise data with mixed `JSON` and `JSON_IETF`
-encodings.
+encodings. Also note, data must be encoded using namespace
+qualified names where required per 
+[RFC7951 Section 4]([https://tools.ietf.org/html/rfc7951](https://datatracker.ietf.org/doc/html/rfc7951#section-6.8)).
 
 Both the client and target MUST support the `JSON` encoding as a minimum.
 


rpc/gnmi/gnmi-specification.md
```diff
@@ -479,7 +479,9 @@ used. This results in quoted string values, and unquoted number values.
 described in [RFC7951](https://tools.ietf.org/html/rfc7951). Data specified with
 a type of JSON MUST be valid JSON, but no additional constraints are placed upon
 it. An implementation MUST NOT serialise data with mixed `JSON` and `JSON_IETF`
-encodings.
+encodings. Also note, data must be encoded using namespace
+qualified names where required per 
+[RFC7951 Section 4]([https://tools.ietf.org/html/rfc7951](https://datatracker.ietf.org/doc/html/rfc7951#section-6.8)).
```
earies, line 484: Agree but my only comment would be - is this necessary?

The encoding is `JSON_IETF` and that encoding should just dictate conformance to RFC7951.  Is there any case where RFC compliance should be diverged here?

It's been a while since I read this entire section (will go back and read) but if it isn't already we should cover what the expectation is for prefix encoding as well w/ this value encoding when it comes to RFC7951 compliance

dplore: Strictly speaking, this reminder in the doc should not be necessary since RFC7951 requires it and OC requires  RFC7951.  But it's observed there is not compliance with this, even within OpenConfig's own published tools.  Given the miss, a little redundancy in the docs seems warranted.

FWIW, I did some searching within OpenConfig repositories and do see this open issue for ygot which is still accurate/unresolved: https://github.com/openconfig/ygot/issues/107

dplore: Ah, nevermind, ygot does implement prefixing:
https://github.com/openconfig/ygot/blob/98713a6f0eb19805b91a4fddec2fa6b2c152e71d/ygot/render.go#L1102

dplore:
Reviewed at March 10, 2026 OC Operators meeting. In light that at least ygot implements this and the RFC already requires prefixing, I will close this PR.

